PR #17, opened 2016-03-07
avoid unbound loop

CJentzsch:
by tallying votes during voting and forbid transfer of tokens until end of the vote


DAO.sol
```diff
@@ -82,10 +84,14 @@ contract DAOInterface {
         bool newServiceProvider;
         // Data needed for splitting the DAO
         SplitData[] splitData;
-        // Array holding all votes that have taken place on the proposal
-        Vote[] votes;
-        // Simple mapping to check if a shareholder has already cast a vote
-        mapping (address => bool) voted;
+        // yay
```
chriseth: This deserves a bit more explanation I think :)

LefterisJP: lol :P

DAO.sol
```diff
@@ -281,11 +281,22 @@ contract DAO is DAOInterface, Token, TokenSale {
 
     function vote(uint _proposalID, bool _supportsProposal) onlyTokenholders noEther returns (uint _voteID) {
         Proposal p = proposals[_proposalID];
-        if (p.voted[msg.sender] || now >= p.votingDeadline) throw;
```
chriseth: `if (_proposalID == 0) throw;`

DAO.sol
```diff
@@ -483,8 +487,11 @@ contract DAO is DAOInterface, Token, TokenSale {
     }
 
 
-    function numberOfVotes(uint _proposalID) constant returns (uint _numberOfVotes) {
-        return proposals[_proposalID].votes.length;
+    function unblock() {
```
chriseth: Automatically call that function as part of `transfer` and `transferFrom`?

DAO.sol
CJentzsch: Because of https://github.com/slockit/DAO/pull/17/files#diff-b30f9d3be2a832949bdd72e23956c130R224 there can never be a proposal with 'ID = 0' . Therefore the ' votingDeadline = 0' and it throws.

DAO.sol
CJentzsch: You would have to call it for every address and this gets you back to a loop ....

DAO.sol
```diff
@@ -281,11 +281,22 @@ contract DAO is DAOInterface, Token, TokenSale {
...
+            p.votedYes[msg.sender] = true;
+        }
+        else{
+            p.nay += balances[msg.sender];
+            p.votedNo[msg.sender] = true;
+        }
+
+        if (blocked[msg.sender] == 0)
+            blocked[msg.sender] = _proposalID;
+        else if (p.votingDeadline > proposals[blocked[msg.sender]].votingDeadline)
```
LefterisJP: I understand the logic here, but maybe add a comment for the `else if` like below:
`If token holder is already blocked by another proposal, but this proposal lasts longer, make this the blocking proposal for the token holder`

DAO.sol
LefterisJP: I also have the same question. Who calls `unblock()` at the moment?

CJentzsch: add the moment the token Holder, but I change it.

LefterisJP:
I left some comments. Also don't forget to run the tests, we got a proposal scenario that tests this code.

DAO.sol
```diff
@@ -483,8 +488,15 @@ contract DAO is DAOInterface, Token, TokenSale {
     }
 
 
-    function numberOfVotes(uint _proposalID) constant returns (uint _numberOfVotes) {
-        return proposals[_proposalID].votes.length;
+    function isBlocked(address _account) returns (bool){
```
chriseth: From the name, this function sounds as if it were side-effect free, which it is not.

PR #19, opened 2016-03-08
Updated comments and complied to 80 Char a line,

GriffGreen:
Please give me feed back on how I shortened line 60-61 and 68-69

### DIFF
--- a/TokenSale.sol
+++ b/TokenSale.sol
@@ -57,16 +57,15 @@ contract TokenSaleInterface {
 
 
 contract TokenSale is TokenSaleInterface, Token {
-    function TokenSale(uint _minValue, uint _closingTime, address _privateSale) 
-    {
+    function TokenSale(uint _minValue, uint _closingTime, address _privateSale) {
         closingTime = _closingTime;
         minValue = _minValue;
         privateSale = _privateSale;
     }
 
     function buyTokenProxy(address _tokenHolder) returns (bool success) {
-        if (now < closingTime && msg.value > 0 && (privateSale == 0 
-        || privateSale == msg.sender)) {
+        if (now < closingTime && msg.value > 0 
+        && (privateSale == 0 || privateSale == msg.sender)) {
             uint token = msg.value;
             balances[_tokenHolder] += token;
             totalSupply += token;
@@ -83,7 +82,7 @@ contract TokenSale is TokenSaleInterface, Token {
     function refund() noEther {
         if (now > closingTime
             && !isFunded
-            && msg.sender.send(balances[msg.sender])) // Execute refund
+            && msg.sender.send(balances[msg.sender])) // execute refund
         {
             Refund(msg.sender, balances[msg.sender]);
             totalSupply -= balances[msg.sender];
--- a/TokenSale.sol
+++ b/TokenSale.sol
@@ -17,28 +17,37 @@ along with the DAO.  If not, see <http://www.gnu.org/licenses/>.
 
 
 /*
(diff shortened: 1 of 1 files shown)


TokenSale.sol
```diff
@@ -48,14 +57,16 @@ contract TokenSaleInterface {
...
+    function TokenSale(uint _minValue, uint _closingTime, address _privateSale) 
+    {
         closingTime = _closingTime;
         minValue = _minValue;
         privateSale = _privateSale;
     }
 
     function buyTokenProxy(address _tokenHolder) returns (bool success) {
-        if (now < closingTime && msg.value > 0 && (privateSale == 0 || privateSale == msg.sender)) {
+        if (now < closingTime && msg.value > 0 && (privateSale == 0
```
LefterisJP: If you split an if then put each conditional in its own line. But in this case since the second condition is an `OR` parenthesized combination please move the entire inner parentheses to the second line.

TokenSale.sol
```diff
@@ -72,7 +83,7 @@ contract TokenSale is TokenSaleInterface, Token {
     function refund() noEther {
         if (now > closingTime
             && !isFunded
-            && msg.sender.send(balances[msg.sender])) // execute refund
+            && msg.sender.send(balances[msg.sender])) // Execute refund
```
LefterisJP: for inline comments (so not comments that describe a function) it's really an overkill to have to capitalize. These are comments for the developers.

TokenSale.sol
```diff
@@ -48,14 +57,16 @@ contract TokenSaleInterface {
 
 
 contract TokenSale is TokenSaleInterface, Token {
-    function TokenSale(uint _minValue, uint _closingTime, address _privateSale) {
+    function TokenSale(uint _minValue, uint _closingTime, address _privateSale)
```
LefterisJP: This goes agains the style guide. In this case, if it's only 1 char above the limit I would not bother (till we have automated tools that do it for us).

But normally the way you would do this depends on style. One way to break it is:

```
    function TokenSale(
        uint _minValue,
        uint _closingTime,
        address _privateSale
) {
```

CJentzsch:
@GriffGreen please check the comments from Lefteris and rebase, can't merge it currently.

TokenSale.sol
```diff
@@ -17,28 +17,37 @@ along with the DAO.  If not, see <http://www.gnu.org/licenses/>.
...
+    bool public isFunded;   
+    // For DAO splits - if privateSale is 0, then it is a public sale, otherwise
+    // only the address stored in privateSale is allowed to purchase tokens
+    address public privateSale;               
+    
+    /// @dev Constructor setting the minimum funding goal and the 
+    /// end of the Token Sale
     /// @param _minValue Token Sale minimum funding goal
     /// @param _closingTime Date (in Unix time) of the end of the Token Sale
+    // This function will not likely be used so it is commented out
```
CJentzsch: This sentence is false. This is the constructor, and since I can not overload it in the implementation, I need to comment it out. But it is used!

TokenSale.sol
```diff
@@ -17,28 +17,37 @@ along with the DAO.  If not, see <http://www.gnu.org/licenses/>.
...
     //  function TokenSale(uint _minValue, uint _closingTime);
 
-    /// @notice buy Token with `_tokenHolder` as the address of the Token holder.
+    /// @notice Buy Token with `_tokenHolder` as the address to send the Token
```
CJentzsch: the tokens gets created here, not send. Sending implies a sender, and there is none. You can leave the comment as is.

PR #28, opened 2016-03-11
Updated indentations

GriffGreen:


DAO.sol
```diff
@@ -296,10 +296,8 @@ contract DAO is DAOInterface, Token, TokenSale {
     // Modifier that allows only shareholders to vote and create new proposals
     modifier onlyTokenholders {
         if (balanceOf(msg.sender) == 0) throw;
-            _
```
CJentzsch: the `_` is very important, it doesn't work and compile without it.

GriffGreen: Oh! Sorry I thought I added that accidentally!

DAO.sol
```diff
@@ -296,10 +296,9 @@ contract DAO is DAOInterface, Token, TokenSale {
...
+            _
```
CJentzsch: please remove the spaces after the `_`   :-)

PR #29, opened 2016-03-11
add price increase during tokensale

CJentzsch:
extra ether are transferred to another account. They can be send to the DAO after they have spent at least this amount of ether.

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -318,6 +318,13 @@ contract DAO is DAOInterface, Token, TokenSale {
         if (address(rewardAccount) == 0) throw;
     }
 
+    function () returns (bool success) {
+        if (now < closingTime + 40 days)
+            return buyTokenProxy(msg.sender);
+        else
+            return receiveEther();
+    }
+
 
     function payDAO() returns (bool) {
         rewards += msg.value;
@@ -657,6 +664,8 @@ contract DAO is DAOInterface, Token, TokenSale {
         if (_recipient == serviceProvider 
             || _recipient == address(rewardAccount) 
             || _recipient == address(this)
+            || (_recipient == address(extraBalance)
+                && totalRewardToken > extraBalance.accumulatedInput())
         )
             return true;
         for (uint i = 0; i < allowedRecipients.length; ++i) {
--- a/DAO.sol
+++ b/DAO.sol
@@ -665,6 +665,8 @@ contract DAO is DAOInterface, Token, TokenSale {
             || _recipient == address(rewardAccount) 
             || _recipient == address(this)
             || (_recipient == address(extraBalance)
+                // only allowed when at least the amount hold in the
+                // extraBalance account has been spent from the DAO
                 && totalRewardToken > extraBalance.accumulatedInput())
         )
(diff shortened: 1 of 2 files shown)


TokenSale.sol
```diff
@@ -85,12 +91,24 @@ contract TokenSale is TokenSaleInterface, Token {
...
+            if (msg.sender.call.value(weiGiven[msg.sender])()) {
+                Refund(msg.sender, weiGiven[msg.sender]);
+                totalSupply -= balances[msg.sender];
+                balances[msg.sender] = 0;
+            }
         }
     }
+
+    function divisor() returns (uint divisor){
+        if (closingTime - 2 weeks > now) return 20;
```
LefterisJP: Definitely add comments here in all 3 legs of the if explaining the different periods and the pricing increase.

TokenSale.sol
```diff
@@ -85,12 +91,24 @@ contract TokenSale is TokenSaleInterface, Token {
 
     function refund() noEther {
         if (now > closingTime
-            && !isFunded
-            && msg.sender.call.value(balances[msg.sender])()) // execute refund
+            && !isFunded)
```
LefterisJP: you can now put the if back in one line :), no longer exceeds the 80 column limit

TokenSale.sol
```diff
@@ -33,7 +33,11 @@ contract TokenSaleInterface {
     bool public isFunded;   
     // For DAO splits - if privateSale is 0, then it is a public sale, otherwise
     // only the address stored in privateSale is allowed to purchase tokens
-    address public privateSale;               
+    address public privateSale;
+    // used to store extra ether which has been paid after the price has increased
```
LefterisJP: after the DAO *token* price has increased

DAO.sol
```diff
@@ -657,6 +664,8 @@ contract DAO is DAOInterface, Token, TokenSale {
         if (_recipient == serviceProvider 
             || _recipient == address(rewardAccount) 
             || _recipient == address(this)
+            || (_recipient == address(extraBalance)
```
LefterisJP: Please write a comment here (on top, not next to it ^_^ ) explaining that the DAO can only receive ether from the extraBalance account only once they have already spent as much as that account's balance is.

CJentzsch: done

DAO.sol
```diff
@@ -657,6 +664,10 @@ contract DAO is DAOInterface, Token, TokenSale {
...
+                // only allowed when at least the amount hold in the
```
LefterisJP: hold -> held

TokenSale.sol
```diff
@@ -84,13 +90,29 @@ contract TokenSale is TokenSaleInterface, Token {
     }
 
     function refund() noEther {
-        if (now > closingTime
-            && !isFunded
-            && msg.sender.call.value(balances[msg.sender])()) // execute refund
+        if (now > closingTime && !isFunded)
```
LefterisJP: also move the `{` up at the right of the parentheses

PR #32, opened 2016-03-15
Homesteadify tests and add insufficient gas at split test

LefterisJP:
- Tests genesis block will now start right after the homestead one 
- Adding test for insufficient gas at DAO split.
   
This test makes sure that if there is not enough gas given at a daoSplit() event the new DAO is not created at all.

This should work thanks to this homestead change:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2.mediawiki#specification

Before homestead the test would fail, but create an empty/invalid new
DAO address which would lead to people burning their tokens in their old
DAO and transferring them to a new invalid DAO, thus losing their money.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -124,6 +124,25 @@ def compare_values(a, b):
 
 
 def eval_test(name, output, expected_dict):
+    """
+    Evaluate output of a scenario and compare with expected results
+        Parameters
+        ----------
+        name : string
+        The name of the scenario to evaluate
+
+        output : string
+        The output of the script that was executed, from which we will
+        extract the results
+
+        expected_dict : dict
+        A dictionary containing all the expected output from the test
+
+        Returns
+        ----------
+        results : dict
+        The dictionary that resulted from the parsing of the test output
+    """
     tests_fail = False
     results = extract_test_dict(name, output)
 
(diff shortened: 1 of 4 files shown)


tests/utils.py
```diff
@@ -157,11 +178,18 @@ def create_genesis(accounts):
     genesis = {}
     genesis["nonce"] = "0xdeadbeefdeadbeef"
     genesis["timestamp"] = "0x0"
-    genesis["parentHash"] = "0x0000000000000000000000000000000000000000000000000000000000000000"
+    # Start after homesteam
+    genesis["parentHash"] = (
+        "0x584bdb5d4e74fe97f5a5222b533fe1322fd0b6ad3eb03f02c3221984e2c0b430"
```
CJentzsch: is this enough? Don't you have to set the blockNumber somewhere?

tests/utils.py
```diff
@@ -157,11 +178,18 @@ def create_genesis(accounts):
...
+    )
     genesis["extraData"] = "0x0"
     genesis["gasLimit"] = "0x8000000"
```
CJentzsch: ' 0x47e7c4' is the current BlockGasLimit. Better use this one for all tests.

tests/utils.py
LefterisJP: Not that I am aware of. Do you know of any other field in the genesis block?

tests/utils.py
LefterisJP: will do

tests/utils.py
CJentzsch: every block has a blockNumber. Who dies he find the block with paretnHash `0x584bdb5d4e74fe97f5a5222b533fe1322fd0b6ad3eb03f02c3221984e2c0b430`
where did you get this hash from?

LefterisJP: You are right, I miscalculated as far as the way of calculating the block hash is concerned. This won't work.

So at the moment there is no way to start from a specific block. Already pinged the GO people to add this feature, or at least specify the block where the homestead change happens.

PR #35, opened 2016-03-16
DAO test framework modularization

LefterisJP:
Each scenario is now in its own directory.

Each directory for now only contains the `run()` scenario python file and the corresponding js script template.


CJentzsch:
merge conflicts

LefterisJP:
rebased

PR #47, opened 2016-03-23
fix minQuorum

CJentzsch:
according to white paper.
Since the amount of ether stored in the DAO can potentially be larger than the number of tokens,


LefterisJP:
Looking good.

PR #48, opened 2016-03-23
use payDAO as fallback

CJentzsch:
to account for rewards from mother DAO counted correctly.
Lets say somone split two time. So we have DAOs A  -B -C (B originates from A, C originates from B). Then C does not receive his reards from A, since they are recieved using the fallback function, and `rewards` is not increased.

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -325,7 +325,7 @@ contract DAO is DAOInterface, Token, TokenSale {
         if (now < closingTime + 40 days)
             return buyTokenProxy(msg.sender);
         else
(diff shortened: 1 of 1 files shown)


LefterisJP:
Okay looking good. To recap. A user of DAO `C` would still need to call `getMyRewards()` on DAO `A` and on DAO `B` to receive rewards, correct?

CJentzsch:
The user of DAO C  can'T do it directly, but when ever B is getting its rewards, he can get it from B.

PR #51, opened 2016-03-25
Make contract updateable per vote

CJentzsch:


LefterisJP:
For a cleaner history you can simply rebase on top of master before merging, so that all those merges don't appear there.

DAO.sol
```diff
@@ -227,6 +227,13 @@ contract DAOInterface {
         address _newServiceProvider
     ) returns (bool _success);
 
+    /// @dev can only be called by the DAO itself through an proposal
```
LefterisJP:
```diff
-an proposal
+ a proposal
```

DAO.sol
```diff
@@ -227,6 +227,13 @@ contract DAOInterface {
...
+    /// updates the contract of the DAO by sending all funds and rewardToken
```
LefterisJP:
```diff
-rewardToken
+rewardTokens
```

Issue #60, opened 2016-03-30
proposalDeposit should be set initatlly to 0


colm:
It is currently 20. This is a problem because in a few years ETH may be worth 1000 USD and in the case where someone wants to withdraw their ETH they will split the DAO and the proposal to withdraw money will require 20*1000 = 20,000 USD. Most people do not have this ETH available and will thus not be able to withdraw it.

colm:
https://github.com/slockit/DAO/pull/61

CJentzsch:
(see comment to the PR)

PR #61, opened 2016-03-30
set inital proposalDeposit = 0 ether

colm:
solves 
https://github.com/slockit/DAO/issues/60

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -319,7 +319,7 @@ contract DAO is DAOInterface, Token, TokenSale {
 
         serviceProvider = _defaultServiceProvider;
         daoCreator = _daoCreator;
-        proposalDeposit = 20 ether;
+        proposalDeposit = 0 ether;
         rewardAccount = new ManagedAccount(address(this));
         if (address(rewardAccount) == 0)
             throw;


CJentzsch:
The initial value doesn't matter much, as long as it is non-zero to provide spam.
The DAO can change this value at any time, by making a proposal to call this function: https://github.com/slockit/DAO/blob/master/DAO.sol#L676

colm:
But in order to pass that proposal someone has to put forward 20 ETH deposit. Correct?

Imagine a future where 1 ETH is 1,000 USD. Someone tries to split the DAO and withdraw their ETH. The new DAO has an initial proposal deposit of 20ETH tehy don't have 20,000 USD  so they cannot make a proposal to change the proposal deposit.

LefterisJP:
As per our discussion in the slack I believe that issue can be up for discussion.
The code here is for a "generic" DAO which will act as a template for all others.
As mentioned in the chat, when splitting it can be a real problem if the person has not enough ether to make a proposal.

I can see 2 solutions here:

1) **Short-term**: Change it to much lower value, but not zero so as to still (kind of) discourage spam. Say 0.01 ETH. The _serious_ DAOs can then make proposals to change their deposit values to more appropriate levels to totally stop spam.

2) **Long-term** Have more special types of proposals. For example detect if the DAO just split out and then make the first proposal deposit-free. Or even make the proposal-deposit change proposals free of deposit. There are a lot of possible options here, but the gist of it is, to have adjustable deposit depending on proposal type.

LefterisJP:
Or even a more elegant short term solution would be:

3) make the proposal deposit a constructor argument so that the initial DAO deployer decides on it. Plus give a default value of `0` when creating a DAO from a `DAOCreator` during a spit so that split DAOs have no artificial caps that make assumptions about ether market value.

colm:
I like no 3. Seems to be best of both worlds with very little added complexity.

PR #62, opened 2016-03-30
removed all p.proposalDeposit from splitDAO

colm:


CJentzsch:
see comment to previous PR

Issue #63, opened 2016-03-30
Split DAO should not have any time for crowd sourceing


colm:
If i am correct and this function is used to create the new split DAO then there should be no + 42 days for crowd sourcing. 

`
    function createNewDAO(address _newServiceProvider) internal returns (DAO _newDAO) {
        NewServiceProvider(_newServiceProvider);
        return daoCreator.createDAO(_newServiceProvider, 0, now + 42 days);
    }
`
Creating a new DAO with 42 days of crowd sourcing time would allow anyone to buy into it and then get a share in the original splitters profits. This should be now + 0 days. It will also result in the new DAO not being able to make any proposals for this time.

CJentzsch:
With `0` , there could only be exactly one who buys into it.
The problem you mention is solved by this: https://github.com/slockit/DAO/blob/master/TokenSale.sol#L78
`privateSale` is set to `0` , when it is public, or in the case of a split, to the address of the Mother DAO. This ways it only accepts funding though the `splitDAO` function. So no one else can participate in the funding period.

colm:
So this means that someone can transfer their ETH anytime during the 42 days. During this time does the new DAO accept proposals? Or does it wait until after the 42 days have elapsed.

CJentzsch:
It waits until the 42 days are over.

PR #64, opened 2016-03-30
set time for crowdsource on split to 0

colm:
https://github.com/slockit/DAO/issues/63


CJentzsch:
this would make it impossible for a group of people to spit.

PR #67, opened 2016-04-05
Add scripts for deploy

colm:
This forums a basis that will make deploying the DAO under various configurations trivial. 

I cannot get this script to deploy the DAO. GIves the following error

`Error: Contract transaction couldn't be found after 50 blocks at DAOCreator creation!`

This is a gas problem that I have not been able to figure out before i had to go offline.


LefterisJP:
I will make some changes by making a PR to your PR and explain why I did what I did. I am not sure if it's a good idea to have deployment scripts inside the DAO framework but let's at least get them fixed.

One first note: Any python code that is written should adhere to the python coding standard. I don't know what OS you have but obtain and run the flake8 executable on any python code you run (https://pypi.python.org/pypi/flake8) and it will tell you which lines are not styled correctly.

LefterisJP:
code.js should not be commited, it's a generated file which will be different each time

LefterisJP:
so here is the PR to the PR with some additions/fixes: https://github.com/colm/DAO/pull/1

LefterisJP:
Danke

Issue #65, opened 2016-03-31
SplitDAO_fundsToBeMoved: Using proposal deposit to increase this.balance, increaseing fundsToBeMoved


colm:
1.  I create a proposal with a huge deposit say 1000 ETH
2.  I split the DAO Which this sets `p.splitData[0].splitBalance = this.balance` effectively. this.balance includes my 1000 ETH

`fundsToBeMoved = (balances[msg.sender] * p.splitData[0].splitBalance) / p.splitData[0].totalSupply;`

4. I get 1000 times my original ETH.
5. My proposal meets the quorum and i get my ETH deposit back

Is this where proposal deposits are stored?

CJentzsch:
you forgot the last part of the equation: `p.splitData[0].splitBalance = this.balance - p.proposalDeposit;`
https://github.com/slockit/DAO/blob/master/DAO.sol#L550

The proposal deposit is not counted in the split balance.

colm:
But `p.proposalDeposit`only contains the proposal deposit for this proposal. If I include the high proposal deposit in another proposal this will not be subtracted out here. Is that correct?

CJentzsch:
correct

colm:
Then this is a bug. I can send a proposal with a huge_deposit. Then create a seperate proposal to Split the DAO get myshares*eth_balance + myshares*huge_deposit. Then when the proposal meets the quoram I get my huge_deposit back.

PR #73, opened 2016-04-08
fix proposal deposit attack

CJentzsch:
fixes https://github.com/slockit/DAO/issues/65


DAO.sol
```diff
@@ -548,7 +556,7 @@ contract DAO is DAOInterface, Token, TokenSale {
             // p.proposalDeposit should be zero here
             if (this.balance < p.proposalDeposit)
```
pirapira: This overflow check should use `sumOfProposalDeposits` instead of `p.proposalDeposit`.

colm:
I believe this opens the door to a 51% attacker preventing splits. With the following actions

1. The attacker creates an ordinary proposal with a deposit that is equal to the current `this.balance`
2. The attacker prevents a quorum being met on this proposal. 
3. When anyone tries to split their new ETH balance will be zero 
`p.splitData[0].splitBalance = this.balance - sumOfProposalDeposits;`
4. The attacker can create a proposal to send the proposal deposit to an address they choose. If the deposit is moved tis way `sumOfProposalDeposits` will never be decremented. Effectively preventing all splits for a temporary deposit.

LefterisJP:
But @colm would not step (1) effectively double `this.balance`?

colm:
Oh yes it would. So let me adjust the attack a little. 

1. creates proposal which huge deposit
2. prevent quoram 
3. Create a proposal to send the amount of the `proposalDeposit` to its service provider. This will decrease `this.balance` but `sumOfProposalDeposits` would remain unchanged. 
4. Splits are prevented. 

If the attacker has trouble getting a quoram for such a large amount of ETH. They can do this 1,000 times with a proposal deposit 1,000 times smaller.

LefterisJP:
> Create a proposal to send the amount of the proposalDeposit to its service provider. This will decrease this.balance but sumOfProposalDeposits would remain unchanged.

No it would not. It would only decrease `this.balance` once the proposal is executed. So once the debating period is over and he calls `executeProposal()`.

CJentzsch:
@colm you can only spend so much ether, that the proposalDeposit remains
https://github.com/slockit/DAO/pull/73/files#diff-b30f9d3be2a832949bdd72e23956c130R498

And, after the proposal debating period ends, everyon can call `executeProposal` and the umOfProposalDeposits` will be decreased, independent of quourum:
https://github.com/slockit/DAO/pull/73/files#diff-b30f9d3be2a832949bdd72e23956c130R522

colm:
Oh yes thats fine then :)

colm:
Should we subtract `sumOfProposalDeposits` [here](https://github.com/slockit/DAO/blob/master/DAO.sol#L693) and [here](https://github.com/slockit/DAO/blob/master/DAO.sol#L722) 

To protect against setting the proposal deposit greater than 10% of the eth balance and abuse of the `minQuorum` formula.

LefterisJP:
@colm yes I believe you are right. Let me handle that.

pirapira:
It might be simpler to store the deposits in a separate managed account.  Then no subtraction is necessary.

LefterisJP:
@pirapira Not a bad idea. Adding yet another ManagedAccount could potentially reduce code readability though.

For the moment I added a function to abstract this whole thing away and avoid code repetition.
@CJentzsch what do you think?

CJentzsch:
@pirapira Seperate account is good. But then the DAO constantly would need to grab the ether from there for proposal which have not reached the quorum. and I actually think it would add complexity.

Issue #78, opened 2016-04-10
calling `newContract()` to defeat the Quorum rule


colm:
Defeat the quorum rules by calling `newContract()` with a users address sending a little gas to this contract and withdrawing all the ETH. 

`totalSupply / minQuorumDivisor+(_value * totalSupply) / (3 * (actualBalance() + totalRewardToken));1`

Here where `_value` should equal `totalSupply` as it is all ETH that is being moved. It will only equal the gas of the transaction. Thus what used to be a 51% attack becomes a 20% attack.

LefterisJP:
Yes this is potentially really dangerous, because as you say you can essentially move the whole balance of the DAO with only 20% quorum.

The only saving grace is that:
```
if (msg.sender != address(this) || !allowedRecipients[_newContract]) return;
```
makes sure that the recipient address is in the whitelist, so you can't just simply send it to just any user or yourself.

On the other hand if you collaborate with a contractor that is already in the whitelist and if you own at least 20% of the tokens you can rob the DAO of all funds.

LefterisJP:
@colm The only way I see solving this (and any other attacks where a party of malicious token holders cooperates with a whitelist member) is to add the ability to remove an allowed address by implementing a `removeAllowedAddress()`.

This way  the curator can remove people who are attacking the DAO from the whitelist and effectively stop the attack.

colm:
I think that most importantly the quorum should never be bi-passed. `removeAllowedAddress()` may be useful for the guardian. But may open the door to abuse. But having addresses permently allowed does seem a little dangerous.

This attack is mitigated by the `splitDAO` and once the quorum abuse is prevented then it will just be a split case.

LefterisJP:
@colm yes all these attacks can be solved by splitting, but the problem here is that the lazy voters are no longer protected in any of these cases. If they don't get enough time to split they will lose all their money to the malicious whitelist member.

colm:
Got to look after the lazy voters. That is the responsibility of the guardian. So I would agree that a guardian needs a `removeAllowedAddress()` in some cases an malicious actor may end up on the whitelist. Thus they will need to be removed.

To what extent should we protect non-malicious white list members from a malicious guardian?

CJentzsch:
I added a the option for the guardians/curators to remove a address from the whitelist.
So in the case the curators are malicous *and* 51% of the token holder. Than you can't do anything to protect the apethetic voter.

CJentzsch:
[...]

PR #79, opened 2016-04-10
Allow removal of malicious whitelist members

LefterisJP:
- Add a function that will allow the guardian (soon to be renamed to
  curator) to remove addresses from the whitelist.

- Assert that the recipient of a proposal is in the whitelist at
  `executeProposal()`

The thinking behind this patch is that a party of malicious token
holders can cooperate with a whitelist member and rob the DAO of its
money.

The token holders can still split away, but those who don't manage to
split in time will stay with a powerless guardian (curator) who will be
unable to stop the proposal from executing. So in this case the lazy
voters would lose all of their money.

By allowing the removal of malicious actors from the whitelist and by
double-checking that someone is still on the whitelist at the time of
`executeProposal()` the guardian (curator) can still save the lazy voters.

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -244,8 +244,14 @@ contract DAOInterface {
     /// @notice Add a new possible recipient `_recipient` to the whitelist so
     /// that the DAO can send transactions to them (using proposals)
     /// @param _recipient New recipient address
-    /// @dev Can only be called by the current service provider
+    /// @dev Can only be called by the current guardian
     function addAllowedAddress(address _recipient) external returns (bool _success);
+
+    /// @notice Remove a `_malicious` recipient from the whitelist so that
+    /// it can no longer receive money from the DAO.
+    /// @param _malicious recipient address for removal
+    /// @dev Can only be called by the current guardian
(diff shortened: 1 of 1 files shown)


LefterisJP:
Related to #78

LefterisJP:
A note about the function name. @GriffGreen proposed during a mumble session that adding/removing people from the whitelist could be a single function with a boolean argument. But I think having 2 separate functions is more user friendly and intuitive. It's open for discussion though.

DAO.sol
```diff
@@ -484,6 +490,8 @@ contract DAO is DAOInterface, Token, TokenSale {
         if (now < p.votingDeadline  // has the voting deadline arrived?
             // Have the votes been counted?
             || !p.open
+            // is the recipient still on the whitelist?
+            || isRecipientAllowed(p.recipient)
```
CJentzsch: so in this case, the voters would be blocked forever and the proposal deposit could never be returned.

CJentzsch:
replaced by https://github.com/slockit/DAO/pull/84/files

PR #82, opened 2016-04-10
Don't close split proposal at executeProposal()

### LINKED ISSUE #81: Prventing splits by `executeProposal` and then unblocking users `isBlocked`
Someone can execute a proposal being used to [split the DAO](https://github.com/slockit/DAO/blob/master/DAO.sol#L551) 

That will close the proposal [here](https://github.com/slockit/DAO/blob/master/DAO.sol#L493) then if an attacker calls `isBlocked` using the spitting user address their `isBlocked` Id will be set to 0 because the proposal is closed. Then [`blocked[msg.sender] != _proposalID`](https://github.com/slockit/DAO/blob/master/DAO.sol#L551) will prevent this user from splitting.

LefterisJP:
Should close #81 

Also @CJentzsch perhaps I should make `isBlocked()` internal? Why allow people externally to invoke the function?

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -491,7 +491,6 @@ contract DAO is DAOInterface, Token, TokenSale {
         }
 
         if (p.newServiceProvider) {
-            p.open = false;
             return;
         }
 


CJentzsch:
replaced by https://github.com/slockit/DAO/pull/84/files

PR #83, opened 2016-04-10
Check call() at newContract() before moving tokens

LefterisJP:
The `call()` to transfer all funds to a new contract could potentially
return an error (eg: stack depth limit reached) and not work, but
without a check the code would happily continue and burn the tokens.

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -607,7 +607,9 @@ contract DAO is DAOInterface, Token, TokenSale {
     function newContract(address _newContract){
         if (msg.sender != address(this) || !allowedRecipients[_newContract]) return;
         // move all funds
-        _newContract.call.value(address(this).balance)();
+        if (!_newContract.call.value(address(this).balance)()) {
+            throw;
+        }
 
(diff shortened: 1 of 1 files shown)


DAO.sol
```diff
@@ -607,7 +607,9 @@ contract DAO is DAOInterface, Token, TokenSale {
     function newContract(address _newContract){
         if (msg.sender != address(this) || !allowedRecipients[_newContract]) return;
         // move all funds
-        _newContract.call.value(address(this).balance)();
+        if (!_newContract.call.value(address(this).balance)()) {
+            return;
```
colm: Should we use `throw` like is used [here](https://github.com/slockit/DAO/blob/master/DAO.sol#L510)

LefterisJP: Since nothing really happens so far in this function I think it makes no actual difference and I wanted to go with the same exit as was used 2 lines above.

CJentzsch: as discussed, `throw` is better here.

PR #84, opened 2016-04-11
allow guardians to remove address from the white list

CJentzsch:


DAO.sol
```diff
@@ -728,10 +736,10 @@ contract DAO is DAOInterface, Token, TokenSale {
     }
 
 
-    function addAllowedAddress(address _recipient) noEther external returns (bool _success) {
+    function changeAllowedRecipients(address _recipient, bool _allowed) noEther external returns (bool _success) {
```
LefterisJP: I think from a UI perspective having 2 different functions may be more beneficial.

DAO.sol
```diff
@@ -490,14 +490,22 @@ contract DAO is DAOInterface, Token, TokenSale {
             throw;
         }
 
-        if (p.newServiceProvider) {
+        if (p.newServiceProvider && now > p.votingDeadline + 28) {
```
LefterisJP: Also rename the proposal's member `p.newServiceProvider` to `newCurator`.

DAO.sol
```diff
@@ -490,14 +490,22 @@ contract DAO is DAOInterface, Token, TokenSale {
...
             p.open = false;
             return;
         }
```
LefterisJP: I suggest:
```js
if (p.newCurator) {
    if (now > p.votingDeadline + 28) {
        p.open = false;
    } 
    return;
}
```

Because as it stands I can still try to execute a split Proposal and continue further on the code, which I should not be able to.

DAO.sol
CJentzsch: The UI is still going to be build (the autogenerated Mist UI is a temporary workaround). And this is only for the curators, not for the masses.

DAO.sol
```diff
@@ -540,11 +548,11 @@ contract DAO is DAOInterface, Token, TokenSale {
 
         if (now < p.votingDeadline  // has the voting deadline arrived?
             //The request for a split expires 41 days after the voting deadline
```
LefterisJP: Also update the amount of days in the comment.

DAO.sol
LefterisJP: Fair enough, in that case a single function is a much better approach.

LefterisJP:
@CJentzsch: Shouldn't you rename all `curators` -> to `curator`? It can be a multi-sig but it's not always gonna be a multi-sig account, right?

DAO.sol
```diff
@@ -490,14 +490,24 @@ contract DAO is DAOInterface, Token, TokenSale {
...
+            if (now > p.votingDeadline + 28) {
+                p.open = false;
+            }
             return;
         }
 
         if (p.amount > actualBalance())
             throw;
 
+        if (!allowedRecipients[p.recipient]) {
```
LefterisJP: Add a comment above here explaining the extraneous conditions when this will happen. Example:
```
// If the curators removed the recipient from the whitelist, close the proposal
```

Issue #75, opened 2016-04-09
Permanently blocked tokens due to malforumed hash


colm:
If someone votes on a proposal that has a malformed hash. [Then](https://github.com/slockit/DAO/blob/master/DAO.sol#L488) 

`if ... ... ... 
           || p.proposalHash != sha3(p.recipient, p.amount, _transactionData)) {
            throw;
        }` 

will cause their tokens to be blocked forever. Preventing them from splitting or transferring.

LefterisJP:
@colm As we also discussed in the chat, voting in a proposal that is either malformed or is made as trap is dangerous. As you mentioned someone could set an almost infinite `votingDeadline` and whoever votes in the proposal will be locked forever. It's not a logic bug and all users should use due diligence and thoroughly check the contents of each proposal they vote for at the moment.

Nonetheless a mistaken proposal index or a bug in the UI leading to calling `vote()` with the trap proposal ID would be fatal for the token holder. We have to evaluate if it is worth giving a vote/proposal canceling functionality or not.

@CJentzsch thoughts?

CJentzsch:
Here is a max debating period: https://github.com/slockit/DAO/pull/85
The hash is neccessary to check the transactionbyteCode.

PR #85, opened 2016-04-11
set max debating period to 8 weeks

CJentzsch:
because of #75
To prevent someone making a extreme long debating period, and people have blocked tokens forever

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -389,6 +389,9 @@ contract DAO is DAOInterface, Token, TokenSale {
             throw;
         }
 
+        if (_debatingPeriod > 8 weeks)
+            throw;
+
         if (!isFunded
             || now < closingTime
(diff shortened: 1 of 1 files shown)


DAO.sol
```diff
@@ -389,6 +389,9 @@ contract DAO is DAOInterface, Token, TokenSale {
             throw;
         }
 
+        if (_debatingPeriod > 8 weeks)
```
LefterisJP: What purpose does this max debating period serve? Perhaps also add a comment here explaining why?

CJentzsch: because of https://github.com/slockit/DAO/issues/75
To prevent someone making a extreme long debating period, and people have blocked tokens forever,

LefterisJP: Oh okay I see, but this is still one of the "do your due diligence" things and check the proposal before voting.

But certainly having a maximum is sensible.

PR #86, opened 2016-04-11
require 53% for a contract update

CJentzsch:


DAO.sol
```diff
@@ -500,6 +500,26 @@ contract DAO is DAOInterface, Token, TokenSale {
 
         uint quorum = p.yea + p.nay;
 
+        // require 53% for callin newContract
```
LefterisJP: callin -> calling

DAO.sol
```diff
@@ -500,6 +500,26 @@ contract DAO is DAOInterface, Token, TokenSale {
...
+        if (_transactionData.length >= 4 && _transactionData[0] == 0x68
+            && _transactionData[1] == 0x37 && _transactionData[2] == 0xff
+            && _transactionData[3] == 0x1e
+            && quorum < minQuorum(actualBalance() + totalRewardToken)) {
+
+            if (!p.creator.send(p.proposalDeposit))
+                throw;
+
+            sumOfProposalDeposits -= p.proposalDeposit;
```
LefterisJP: abstract the following into a single internal function.
```js
sumOfProposalDeposits -= p.proposalDeposit;
// Since the voting deadline is over, close the proposal
p.open = false;
// Initiate event
ProposalTallied(_proposalID, _success, quorum);
```

Then use it here, and also again at the end of `executeProposal()`

DAO.sol
```diff
@@ -500,6 +500,26 @@ contract DAO is DAOInterface, Token, TokenSale {
 
         uint quorum = p.yea + p.nay;
 
+        // require 53% for callin newContract
+        if (_transactionData.length >= 4 && _transactionData[0] == 0x68
+            && _transactionData[1] == 0x37 && _transactionData[2] == 0xff
```
LefterisJP: funny side effect is that if any other contract on the whitelist ever has the signature: `newContract(address)` it will fail here :)

DAO.sol
```diff
@@ -500,8 +500,19 @@ contract DAO is DAOInterface, Token, TokenSale {
...
+        bool updateContractQuroumCheck = true;
+        if (_transactionData.length >= 4 && _transactionData[0] == 0x68
+            && _transactionData[1] == 0x37 && _transactionData[2] == 0xff
+            && _transactionData[3] == 0x1e
+            && quorum < minQuorum(actualBalance() + totalRewardToken)) {
+
+                updateContractQuroumCheck = false;
+        }
+
+
```
LefterisJP: remove extraneous whitespace from here

DAO.sol
```diff
@@ -500,8 +500,19 @@ contract DAO is DAOInterface, Token, TokenSale {
 
         uint quorum = p.yea + p.nay;
 
+        // require 53% for callin newContract()
```
LefterisJP: still comment should be `calling` and not `callin`

DAO.sol
```diff
@@ -500,8 +500,19 @@ contract DAO is DAOInterface, Token, TokenSale {
...
+        bool updateContractQuroumCheck = true;
```
LefterisJP: Update variable name in all occurences. `updateContractQuroumCheck` -> `updateContractQuorumCheck`

CJentzsch: ?

DAO.sol
```diff
@@ -513,7 +524,7 @@ contract DAO is DAOInterface, Token, TokenSale {
             lastTimeMinQuorumMet = now;
             rewardToken[address(this)] += p.amount;
             totalRewardToken += p.amount;
-        } else if (quorum >= minQuorum(p.amount) && p.nay >= p.yea) {
+        } else if (quorum >= minQuorum(p.amount) && p.nay >= p.yea || !updateContractQuroumCheck) {
```
LefterisJP: What happens if the quorum checks out and `updateContractQuorumCheck = true` and we get here because people downvoted the proposal? Isn't the proposal deposit lost?

CJentzsch: No, proposal deposit only gets lost when less than 20% voted for it.

LefterisJP:
All right, looks good then. But it's 7 commits for 1 small thing so please squash them into one commit before merging.

DAO.sol
```diff
@@ -500,8 +500,18 @@ contract DAO is DAOInterface, Token, TokenSale {
 
         uint quorum = p.yea + p.nay;
 
+        // require 53% for calling newContract()
+        bool updateContractQuorumCheck = true;
+        if (_transactionData.length >= 4 && _transactionData[0] == 0x68
+            && _transactionData[1] == 0x37 && _transactionData[2] == 0xff
+            && _transactionData[3] == 0x1e
```
Sjors: I'm trying to wrap my head around what these checks are doing. Are `0x68`, `0x37`, `0xf`f and `0x1e` some sort of opcodes? Where do I find their meaning?

This code might be easier to read if you put these hex numbers in a named `uint constant`.

LefterisJP: This is the hash signature of `newContract()`. The comment explains what it is quite well.
To obtain it just run `solc --hashes DAO.sol`

PR #88, opened 2016-04-11
Constants and readability

LefterisJP:
- Move all hardcoded values as constants in the top of the contract
- Explicitly set proposals.length in the constructor, it's more readable this way.
- Fixed a timestamp bug. `28` != `28 days`

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -26,6 +26,18 @@ import "./ManagedAccount.sol";
 
 contract DAOInterface {
 
+    // The amount of days for which people who try to participate in the sale
+    // by calling the fallback function will still get their money back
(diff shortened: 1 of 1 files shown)


DAO.sol
```diff
@@ -496,7 +508,7 @@ contract DAO is DAOInterface, Token, TokenSale {
         }
 
         if (p.newCurators) {
-            if (now > p.votingDeadline + 28) {
```
LefterisJP: @CJentzsch I introduce a logic change here, but I think this was a typo so it's safe.

LefterisJP:
@CJentzsch: You used both 28 and 27 days for the splitExecutionPeriod in the code before. What should it be at the end?

PR #94, opened 2016-04-12
Avoid blocked token because of throwing calls

CJentzsch:
The problem is, in the case `executeProposal` throws, all the tokens that have voted are blocked forever (big problem).
Here some scenarios:
1) the fallback function of the creator of the proposal throws ( I would consider it an attack to create a proposal with an address that has a fallback function that can throw)
2) the called recipient functions throws
3) another proposal got accepted in the meantime, and the DAO doesn't has enough funds to execute the proposal

This PR solves 1) by letting the curators close a proposal by removing it from the whitelist.
2) and 3) are solved by not throwing anymore but just closing the proposal.

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -37,7 +37,7 @@ contract DAOInterface {
     uint constant splitExecutionPeriod = 27 days;
     // Period of time after which the minimum Quorum is halved
     uint constant quorumHalvingPeriod = 52 weeks;
-    // Period after which a propsal can be closed (used in the case `executeProposal`
+    // Period after which a proposal can be closed (used in the case `executeProposal`
     // fails because it throws)
     uint constant executeProposalPeriod = 5 days;
 
--- a/DAO.sol
+++ b/DAO.sol
@@ -497,6 +497,7 @@ contract DAO is DAOInterface, Token, TokenSale {
         bytes _transactionData
     ) noEther returns (bool _success) {
 
+        stackDepthControl();
         Proposal p = proposals[_proposalID];
 
         if (p.newCurator) {
@@ -821,6 +822,17 @@ contract DAO is DAOInterface, Token, TokenSale {
     function createNewDAO(address _newCurator) internal returns (DAO _newDAO) {
         NewCurator(_newCurator);
         return daoCreator.createDAO(_newCurator, 0, 0, now + splitExecutionPeriod);
+    }
+
+    uint stackDepthControlTmp;
+
+    function stackDepthControl() returns (bool) {
+        if (stackDepthControlTmp++ < 16){
+            if (!DAO(this).stackDepthControl())
+                throw;
+        }
+        stackDepthControlTmp = 0;
+        return true;
     }
 
 
--- a/DAO.sol
+++ b/DAO.sol
@@ -499,12 +499,18 @@ contract DAO is DAOInterface, Token, TokenSale {
 
         Proposal p = proposals[_proposalID];
 
+        if (p.newCurator) {
+            if (now > p.votingDeadline + splitExecutionPeriod) {
+                p.open = false;
+            }
+            return;
+        }
+
         // If the curator removed the recipient from the whitelist, close the proposal
         // in order to free the deposit and allow unblocking of voters
         if (!allowedRecipients[p.recipient]) {
-            p.open = false;
+            closeProposal(_proposalID);
             p.creator.send(p.proposalDeposit);
-            sumOfProposalDeposits -= p.proposalDeposit;
             return;
         }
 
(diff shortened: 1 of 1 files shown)


DAO.sol
```diff
@@ -515,60 +506,78 @@ contract DAO is DAOInterface, Token, TokenSale {
...
-            // Without this throw, the creator of the proposal can repeat this,
-            // and get so much ether
-            if (!p.recipient.call.value(p.amount)(_transactionData))
-                throw;
+
+            lastTimeMinQuorumMet = now;
+
+            if (!p.recipient.call.value(p.amount)(_transactionData)) {
+                closeProposal(_proposalID);
+                ProposalTallied(_proposalID, _success, quorum);
```
colm: Should we put `lastTimeMinQuorumMet = now; ` here?

DAO.sol
```diff
@@ -814,6 +824,17 @@ contract DAO is DAOInterface, Token, TokenSale {
         return daoCreator.createDAO(_newCurator, 0, 0, now + splitExecutionPeriod);
     }
 
+    uint stackDepthControlTmp;
+
+    function stackDepthControl() returns (bool) {
```
LefterisJP: This will definitely need a comment explaining functionality :)

DAO.sol
```diff
@@ -515,60 +507,78 @@ contract DAO is DAOInterface, Token, TokenSale {
...
         ProposalTallied(_proposalID, _success, quorum);
     }
 
 
+    function closeProposal(uint _proposalID) internal {
+        Proposal p = proposals[_proposalID];
+        sumOfProposalDeposits -= p.proposalDeposit;
+        p.open = false;
+    }
+
```
LefterisJP: remove extraneous whitespace

DAO.sol
```diff
@@ -515,60 +507,78 @@ contract DAO is DAOInterface, Token, TokenSale {
...
+        bool proposalCheck = true;
+
+        if (p.amount > actualBalance())
+            proposalCheck = false;
+
         uint quorum = p.yea + p.nay;
 
         // require 53% for calling newContract()
-        bool updateContractQuorumCheck = true;
+
```
LefterisJP: no need for whitespace here between comment and if statement

DAO.sol
```diff
@@ -814,6 +824,17 @@ contract DAO is DAOInterface, Token, TokenSale {
         return daoCreator.createDAO(_newCurator, 0, 0, now + splitExecutionPeriod);
     }
 
+    uint stackDepthControlTmp;
+
+    function stackDepthControl() returns (bool) {
+        if (stackDepthControlTmp++ < 16){
```
LefterisJP: Since `16` here is a constant that we choose, move it to the top of the interface with a descriptive name as a `uint const` like the others we are gathering there. Possible name:  `stackDepthCeiling`

DAO.sol
CJentzsch: done

DAO.sol
```diff
@@ -515,60 +506,80 @@ contract DAO is DAOInterface, Token, TokenSale {
...
+    function closeProposal(uint _proposalID) internal {
+        Proposal p = proposals[_proposalID];
+        if (p.open)
+            sumOfProposalDeposits -= p.proposalDeposit;
+        p.open = false;
+    }
+
+    function emergenyCloseProposal(uint _proposalID) external {
+        Proposal p = proposals[_proposalID];
+        if (p.votingDeadline + 5 days < now && p.open)
```
LefterisJP: Please move this `5 days` as a `uint constant` in the contract beginning.

DAO.sol
```diff
@@ -515,60 +506,80 @@ contract DAO is DAOInterface, Token, TokenSale {
...
         // If the curator removed the recipient from the whitelist, close the proposal
         // in order to free the deposit and allow unblocking of voters
-        if (!allowedRecipients[p.recipient]) {
-            p.open = false;
-            if (!p.creator.send(p.proposalDeposit))
-                throw;
-            sumOfProposalDeposits -= p.proposalDeposit;
+        if (!allowedRecipients[p.recipient] && p.open) {
+            closeProposal(_proposalID);
+            p.creator.send(p.proposalDeposit);
```
LefterisJP: Should we not have a check to reduce the sumOfProposalDeposits here? Look below:
```js
if(p.creator.send(p.proposalDeposit)) {
    sumOfProposalDeposits -= p.proposalDeposit;
}
```

CJentzsch: whether it is send or not, `sumOfProposalDeposits` get reduced in `closeProposal` .

LefterisJP: Ah yes. I missed that part. 

Then there is another problem. If it's not sent, then the `actualBalance()` will be returning a wrong number. The DAO will have more money than it thinks it does.

CJentzsch: Why?

LefterisJP: Yes you are correct. Disregard my comment.

DAO.sol
```diff
@@ -515,60 +509,80 @@ contract DAO is DAOInterface, Token, TokenSale {
...
 
 
+    function closeProposal(uint _proposalID) internal {
+        Proposal p = proposals[_proposalID];
+        if (p.open)
+            sumOfProposalDeposits -= p.proposalDeposit;
+        p.open = false;
+    }
+
+    function emergenyCloseProposal(uint _proposalID) external {
```
LefterisJP: Add `emergencyCloseProposal()` to the interface with a documentation string

CJentzsch: done

PR #96, opened 2016-04-13
fix minQuorum by using only the rewardToken in this DAO

CJentzsch:
After many splits, `totalRewardToken` can be much higher than the rewardToken owned by this DAO.


DAO.sol
```diff
@@ -406,7 +406,7 @@ contract DAO is DAOInterface, Token, TokenSale {
             || msg.value > 0
             || _debatingPeriod < minSplitDebatePeriod)) {
             throw;
-        } else if(
+        } else if (
```
LefterisJP: extra whitespace :)

PR #103, opened 2016-04-15
better max proposal deposit

CJentzsch:
In the case `actualBalance` is zero, it was not possible to change the `proposalDeposit`

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -40,14 +40,16 @@ contract DAOInterface {
     // Period after which a proposal can be closed
     // (used in the case `executeProposal` fails because it throws)
     uint constant executeProposalPeriod = 5 days;
-
+    // Denotes the maximim proposal deposit that can be given. It is given as
+    // a fraction of total Ether spent plus balance of the DAO
+    uint constant maxDepositDivisor = 100;
     // Proposals to spend the DAO's ether or to choose a new Curator
     Proposal[] public proposals;
     // The quorum needed for each proposal is partially calculated by
     // totalSupply / minQuorumDivisor
-    uint minQuorumDivisor;
+    uint public minQuorumDivisor;
     // The unix time of the last time quorum was reached on a proposal
-    uint lastTimeMinQuorumMet;
+    uint  public lastTimeMinQuorumMet;
 
     // Address of the curator
     address public curator;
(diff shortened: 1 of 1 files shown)


DAO.sol
```diff
@@ -84,7 +84,7 @@ contract DAOInterface {
     uint public proposalDeposit;
 
     // the accumulated sum of all current proposal deposits
-    uint sumOfProposalDeposits;
+    uint public sumOfProposalDeposits;
```
LefterisJP: Was this done on purpose? Or did you make it public only for testing purposes and forgot it there?

DAO.sol
```diff
@@ -782,7 +782,7 @@ contract DAO is DAOInterface, Token, TokenSale {
 
 
     function changeProposalDeposit(uint _proposalDeposit) noEther external {
-        if (msg.sender != address(this) || _proposalDeposit > actualBalance() / 10)
+        if (msg.sender != address(this) || _proposalDeposit > (actualBalance() + rewardToken[address(this)]) / 100)
```
LefterisJP: Maybe make this `100` into a constant and put it in the contract beginning too?
Something like:
```js
    // Denotes the maximim proposal deposit that can be given. It is given as 
    // a fraction of total Ether spent plus balance of the DAO
    uint constant maxDepositDivisor = 100;
```

DAO.sol
CJentzsch: On purpose. Currently it's hard for users to know what is the actual balance. Either we make `actualBalance` non-internal, or this one public. Same result.

LefterisJP: Then perhaps make the `actualBalance()` as external. Since it would give the true balance of the DAO, while with `sumOfProposalDeposits` being `public` the user would need to also do a mental subraction to get that amount.

DAO.sol
```diff
@@ -40,14 +40,16 @@ contract DAOInterface {
     // Period after which a proposal can be closed
     // (used in the case `executeProposal` fails because it throws)
     uint constant executeProposalPeriod = 5 days;
-
+    // Denotes the maximim proposal deposit that can be given. It is given as
```
LefterisJP: `maximim` -> `maximum`  (I had a typo in my comment)

DAO.sol
```diff
@@ -40,14 +40,16 @@ contract DAOInterface {
...
+    // a fraction of total Ether spent plus balance of the DAO
+    uint constant maxDepositDivisor = 100;
     // Proposals to spend the DAO's ether or to choose a new Curator
```
LefterisJP: at least 1 space between the function constants and the rest, it helps visually when reading the contract.

DAO.sol
LefterisJP: ^ What about this? Why no make `actualBalance()` a public function?

PR #104, opened 2016-04-15
log event AllowedRecipientChanged

pirapira:
Before this commit an event `AllowedRecipientAdded` was declared but never used.  Since now allowed recipients can both be added and removed, I changed the name of the event.  And the event is logged.


LefterisJP:
Hehe nicely spotted.

PR #112, opened 2016-04-18
only DAO can receive extraBalance

CJentzsch:
@LefterisJP found an issue, that it is possible for someone having 20% of the token to try to send the ether of the extraBalance to himself instead of the DAO. This PR fixes that.


LefterisJP:
Looks good. Can you add a description of the attack in the PR's text description?

PR #115, opened 2016-04-18
Minimum fueling goal value is denominated in Wei

LefterisJP:
According to the documentation, the minimum fueling goal should
be in Wei and not in number of tokens.

*WARNING* Mutually exclusive with #114. Merge one or the other.


CJentzsch:
This will add a lot of gas costs since this function will be called many many times. Merged the other PR.

LefterisJP:
Yeah I thought so too, just wanted to give you options ;)

PR #117, opened 2016-04-19
Close all types of overdue proposals at executeProposal

LefterisJP:
- Close all types or proposals at `executeProposal()` if their voting
  deadline plus waiting period has expired.

- Remove the now no longer needed `emergenyCloseProposal()`

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -37,9 +37,9 @@ contract DAOInterface {
     uint constant splitExecutionPeriod = 27 days;
     // Period of time after which the minimum Quorum is halved
     uint constant quorumHalvingPeriod = 52 weeks;
-    // Period after which a proposal can be closed
+    // Period after which a proposal is closed
     // (used in the case `executeProposal` fails because it throws)
-    uint constant executeProposalPeriod = 5 days;
+    uint constant executeProposalPeriod = 10 days;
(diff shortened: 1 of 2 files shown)


DAO.sol
```diff
@@ -512,10 +506,12 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
         Proposal p = proposals[_proposalID];
 
-        if (p.newCurator) {
-            if (now > p.votingDeadline + splitExecutionPeriod) {
-                p.open = false;
-            }
+        uint waitPeriod = p.newCurator
+            ? splitExecutionPeriod
+            : executeProposalPeriod;
```
CJentzsch: please increase the `executeProposalPeriod` to 10 days.

DAO.sol
```diff
@@ -512,10 +506,12 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+        // If we are over deadline and waiting period, assert proposal is closed
+        if (now > p.votingDeadline + waitPeriod) {
```
CJentzsch: it also saves some gas if you would add `&& p.open` here. But not neccessary.

CJentzsch:
Please also update the WP (just remove the function there)

PR #126, opened 2016-04-22
add updateClientAddress function

CJentzsch:
in the case the DAO updates it contract, it should also be able to update its address here

### DIFF
--- a/SampleOffer.sol
+++ b/SampleOffer.sol
@@ -118,6 +118,10 @@ contract SampleOffer {
         deploymentReward = _deploymentReward;
     }
 
+    function updateClientAddress(DAO _newClient) callingRestriction {
+        client = _newClient;
+    }
+
     // interface for Ethereum Computer
     function payOneTimeReward() returns(bool) {
         if (msg.value < deploymentReward)


LefterisJP:
Hmm but then this call should be timed very correctly by the DAO's users so that it happens right before a proposal to move to a new contract is passed.

So the process would be like that:

1) Deploy a new DAO contract. Get its address
2) Make  a proposal to call `updateClientAddress(addressOfDAODeployatAtStep1)` function. Once this proposal passes then no further interaction with the contractor is possible.
3) Make a proposal (and pass it) to call `newContract(addressOfDAODeployatAtStep1)`

From step 2 and on all proposals *have to* pass. That's kinda dangerous.

CJentzsch:
No, they can safely do this after they update there contract with `newContract` .
This function "only" moves all the ether and rewardTokens. But they cans till make a proposal an vote for that in order to call `updateClientAddress`.

CJentzsch:
@LefterisJP please explain in the case you still think there is a problem

LefterisJP:
@CJentzsch No I don't believe there is a problem. Still they would have to manage to get a quorum in a vote in the old DAO to change the contractor. I guess at least with the current model there is no way to avoid this.

CJentzsch:
yes. But even if they don't, it wouldn't be a big problem. The rewards would still get forwarded.

PR #127, opened 2016-04-22
avoid waiting until end of grace period to refund the extra balance

CJentzsch:
In the case the min goal is not reached. Currently the ether from the extra balance can not be retrieved until after the grace period is over, this PR fixes this.


TokenCreation.sol
```diff
@@ -112,7 +112,8 @@ contract TokenCreation is TokenCreationInterface, Token {
     function refund() noEther {
         if (now > closingTime && !isFueled) {
             // Get extraBalance - will only succeed when called for the first time
-            extraBalance.payOut(address(this), extraBalance.accumulatedInput());
+            if (extraBalance.balance >= extraBalance.accumulatedInput())
```
LefterisJP: This if check is just to save some gas by not calling a no-effect function again, right?

LefterisJP:
It looks good to me. I wrote a test for this and `extraBalance` related refund and want to add this to the PR before merging. Has some issues that need ironing out. Will submit it sometime tomorrow as early as possible.

LefterisJP:
Added the test which fails to fuel the DAO while the extraBalance account
is in use. Afterwards we check that refund works and the full amount is
given back to the user including their portion of what should be in the
extraBalance.

The test does its job since it fails before commit
9ebf64b44bf5948a2df48a7072168106fcb1f69b but after that commit it is
always succesful.

@CJentzsch You can merge when ready.

PR #131, opened 2016-04-25
improve min quorum rules

CJentzsch:
allow curator to reduce minQuorum + set back min quourum in the case 20% is reached

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -825,7 +825,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
         // this can only be called after `quorumHalvingPeriod` has passed or at anytime
         // by the curator with a delay of at least `minProposalDebatePeriod` between the calls
         if ((lastTimeMinQuorumMet < (now - quorumHalvingPeriod) || msg.sender == curator)
-			&& lastTimeMinQuorumMet < (now + minProposalDebatePeriod)) {
+			&& lastTimeMinQuorumMet < (now - minProposalDebatePeriod)) {
             lastTimeMinQuorumMet = now;
             minQuorumDivisor *= 2;
             return true;
--- a/DAO.sol
+++ b/DAO.sol
@@ -552,13 +552,18 @@ contract DAO is DAOInterface, Token, TokenCreation {
                 proposalCheck = false;
         }
 
-        // Execute result
-        if (quorum >= minQuorum(p.amount) && p.yea > p.nay && proposalCheck) {
+        if (quorum >= minQuorum(p.amount)) {
             if (!p.creator.send(p.proposalDeposit))
                 throw;
 
             lastTimeMinQuorumMet = now;
+            // set the minQuorum to 20% again, in the case it has been reached
+            if (quorum > totalSupply / 5)
+                minQuorumDivisor = 5;
+        }
 
+        // Execute result
+        if (quorum >= minQuorum(p.amount) && p.yea > p.nay && proposalCheck) {
             if (!p.recipient.call.value(p.amount)(_transactionData))
                 throw;
 
@@ -566,10 +571,6 @@ contract DAO is DAOInterface, Token, TokenCreation {
             _success = true;
             rewardToken[address(this)] += p.amount;
             totalRewardToken += p.amount;
-        } else if (quorum >= minQuorum(p.amount) && p.nay >= p.yea || !proposalCheck) {
-            if (!p.creator.send(p.proposalDeposit))
-                throw;
-            lastTimeMinQuorumMet = now;
         }
 
         closeProposal(_proposalID);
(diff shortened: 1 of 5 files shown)


DAO.sol
```diff
@@ -821,7 +821,8 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
 
     function halveMinQuorum() returns (bool _success) {
-        if (lastTimeMinQuorumMet < (now - quorumHalvingPeriod)) {
+        if ((lastTimeMinQuorumMet < (now - quorumHalvingPeriod) || msg.sender == curator)
```
LefterisJP: Add a comment here and in the interface for `halveMinQuorum()` stating that the curator/s can now halve minQuorum in case of an emergency.

DAO.sol
```diff
@@ -821,7 +821,8 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+			&& lastTimeMinQuorumMet < (now + minProposalDebatePeriod)) {
```
LefterisJP: I am not sure I understand the purpose of this condition line. Can you explain it to me and perhaps add a comment for it?

DAO.sol
```diff
@@ -552,24 +552,24 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
         }
 
-        // Execute result
-        if (quorum >= minQuorum(p.amount) && p.yea > p.nay && proposalCheck) {
+        if (quorum >= minQuorum(p.amount)) {
             if (!p.creator.send(p.proposalDeposit))
                 throw;
 
             lastTimeMinQuorumMet = now;
+            if (quorum > totalSupply / 5)
```
LefterisJP: Add a comment here stating that every time quorum reaches a 20% the minQuorumDivisor will be reset back to `5` making it back to 20% from whatever value it was before.

DAO.sol
CJentzsch: this protect the DAO from a malicous Curator, he calls this function again and again, to set it close to 0. Now the min time between calling 
'halveMinQuorum` is 2 weeks. When the DAO sees the attack from the curator, they should split.

LefterisJP: Okay I see. Write that in a comment there since that can't be guessed from the code.

Also if the changes of this PR do make it in the code we would also need to change the whitepaper to include them since they are pretty important.

DAO.sol
```diff
@@ -821,6 +822,8 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
 
     function halveMinQuorum() returns (bool _success) {
+        // this can only be called after `quorumHalvingPeriod` has paased or by the curator
```
LefterisJP: `has paased or by the curator` -> `has passed or at anytime by the curator`

DAO.sol
```diff
@@ -821,6 +822,8 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+        // with a time of at least `minProposalDebatePeriod` between the calls
```
LefterisJP: with a time -> with a delay

CJentzsch:
This is a contoversial PR, needs discussion.

LefterisJP:
@CJentzsch So this PR looks okay to me as far as merging is concerned. Added some gas adjustments in the tests just to be on the safe side with some tests that were failing occasionally.

DAO.sol
```diff
@@ -821,7 +822,10 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
 
     function halveMinQuorum() returns (bool _success) {
-        if (lastTimeMinQuorumMet < (now - quorumHalvingPeriod)) {
+        // this can only be called after `quorumHalvingPeriod` has passed or at anytime
+        // by the curator with a delay of at least `minProposalDebatePeriod` between the calls
+        if ((lastTimeMinQuorumMet < (now - quorumHalvingPeriod) || msg.sender == curator)
+			&& lastTimeMinQuorumMet < (now - minProposalDebatePeriod)) {
```
colm: The `minProposalDebatePeriod` is here to give token holders a chance to split before the quorum gets reduced? 

But the user has no indication of the curators intent to reduce the quorum until they call this function. 

Which can be call after just before a proposal that does not meet the current quorum is executed reducing the quorum and allowing it be executed. Allowing the quorum be reduced second before ETH gets transfered.

CJentzsch: This is only there to prevent the curator calling it hundreds of times within a minute, and then start an attack with low minQourum. On the other hand, if your curator is malicious, you should split anyway, and this gives you the time to do it.

colm: @CJentzsch My problem here is that there is no way to know if the Curator is malicious until all your ETH has been disappeared. The time between the requirement for a malicious action and all your ETH being gone is 15 seconds. 

Consider the following:
1. I create 1,000 proposals to send 1 ETH, I have 15% of DAO. 
2. I vote for my proposals no one else does. It will not meet the quorum. I will lose my deposit
3. At the last second the curator reduces teh quorum. I execute the proposals in the next block. All eth goes to me.

CJentzsch: for 1.) your address has to be on whitelist. If a non-sense address makes it to the whitelist, then this is an opportunity to split.

So the attack is only possible if a malicious address is on the whitelist. In the case, by accident, the curator sets such a n address on the whitelist, and then you see 1000 proposals to him, then the curator should remove him from the list, if he does not, then you have malicious curator, then you have to split.

colm: Well then I could do the same using slockits address and have you guys removed from the list. 

The attack still works if the first part is excluded and there is just 1 proposals. That does not reach the quorum.

Basically i don't like that the quorum can be changed so quickly. which changes a proposals that didn't reach the quorum to one that passed.

PR #132, opened 2016-04-25
add possibility to be a non-voter

CJentzsch:
This is highly controversial. This PR is a suggestion which needs to be discussed ...


LefterisJP:
This will need extensive testing. Will review as soon as I find time.

PR #133, opened 2016-04-26
avoid creating reward tokens when send to DAO itself

CJentzsch:


LefterisJP:
Looks fine to me and it passed the important tests I ran on it.

PR #135, opened 2016-04-27
Force contractor to promise open source

tjayrush:
An attempt to make explicit two things: (1) a promise from the contractor that the project developed under this proposal will be open source, and (2) an expectation (or a lack of an expectation) from the DAO that the materials will be open source.

Note that if the DAO sends 'false' to the sign function, they are saying to do not expect open source (but will accept it if offered). If the DAO sends 'true,' then the contract sill be signed only if the contractor has already promised to develop open source.

Forcing the contractor to develop open source (at the DAO's discretion) is important to insure that the product that is ultimately developed is exactly what the DAO is paying for. Trust but verify.

### DIFF
--- a/SampleOffer.sol
+++ b/SampleOffer.sol
@@ -39,6 +39,7 @@ contract SampleOffer {
     DAO client; // address of DAO
 
     bool public promiseValid;
+    bool public isOpenSource;
     uint public rewardDivisor;
     uint public deploymentReward;
 
@@ -54,6 +55,12 @@ contract SampleOffer {
 
     modifier onlyClient {
         if (msg.sender != address(client))
+            throw;
+        _
+    }
+
+    modifier onlyContractor {
+        if (msg.sender != contractor)
             throw;
         _
     }
@@ -77,14 +84,20 @@ contract SampleOffer {
         deploymentReward = _deploymentReward;
     }
 
-    function sign() {
+    function sign(bool expectsOpenSource) {
         if (msg.value < totalCosts || dateOfSignature != 0)
             throw;
         if (!contractor.send(oneTimeCosts))
             throw;
+        if (expectsOpenSource && !isOpenSource)
+            throw;
         client = DAO(msg.sender);
         dateOfSignature = now;
         promiseValid = true;
+    }
+    
+    function promiseOpenSource() onlyContractor {
+        isOpenSource = true;
     }
 
     function setDailyCosts(uint _dailyCosts) onlyClient {


LefterisJP:
I can see the sentiment behind the PR but this also raises a kind of more generic question. How are statements like this one supposed to be enforced in the real world? We promise to be open source, we promise to reduce our carbon footprint etc.

I don't think that terms like this should be encoded in the smart contract offers because if we go down this way then we should have a variable for any possible combination of terms a contract would have.

Smart contracts, even though the name may suggest otherwise, are not like real world contracts. They are more like black boxes of code which execute whenever poked by outside input.

jeffanthony:
I'm not a pro but it seems like this is already a feature. You could write up your legal document complete with Legal names and Ethereum addresses controlled on both sides and then a hash of the legal document could be included in...

`bytes32 _hashOfTheTerms`

tjayrush:
After I posted the PR, I thought it should have been more generic. Instead of promiseOpenSource, it should be something like senderAgrees(...). In C++, I would make a bit mask with each term being agreed to set to 1. The contractor would send the appropriate bit mask, and the DAO would send the same bit mask to the 'sign' function. The _hashOfTheTerms text would make the bits specific. Not sure if bit masks work in Solidity.

The idea is something that operates like those small spaces on real world contracts where you have to explicitly put your initials. Both parties sign the contract, but both parties also initial certain keys sections or terms of the contracts. It makes the intentions of each party more clear.

I hear you @jeffanthony about the _hashOfTheTerms, but that's "on-paper" natural language stuff. I'm trying to build a mechanism into the source to offload some of those natural language terms into the actual data stored on the blockchain.

I would suggest that the SampleOffer.sol code be removed from the DAO code so it doesn't get deployed (and therefore locked forever) by future generations. The actual SampleOffer.sol code has nothing to do with the actual DAO code anyway. In that way, the community can further discuss before committing forever to the starting point for all future proposals.

But--I won't argue the point. I'm just establishing the point here. I think the _hashOfTheTerms thing will be something that opens up a ton of questions. For example, why store it if the code is king?

christoph2806:
Smart contracts should contain only clauses a smart contract can enforce.

tjayrush:
How exactly does the smart contract enforce _hashOfTheTerms? It stores the data and then literally never uses it again. Plus the proposal I made does enforce itself because the 'sign' function throws if the DAO is asking the contractor to initial the little lines on the page, but the contractor hasn't explicitly done so.

But my question still stands. Does the SampleOffer.sol get deployed in the impending deploy of the DAO? Because if it doesn't, then it's a moot point.

Sorry. I'm new to Github. How does an issue get closed? I guess I made my argument, so it's up to whomever to close it. Thanks for letting me make my point. I'd love to hear what anyone else might think.

christoph2806:
_hashOfTheTerms enforces the Contractor not to tamper or counterfeit the proposal. That's the only function of this. Whether the Contractor adheres to the spirit of the proposal, is not enforcable by the contract. But the contractor can never deny to have offered the proposal with hash _hashOfTheTerms.

jeffanthony:
This would be an interesting discussion for the [daohub forum](https://forum.daohub.org/c/theDAO)

tjayrush:
I'll close this or now because I think if it were to be added, it should be expanded to enable a broader range of "terms." Thanks for the feedback on my first official PR.

PR #136, opened 2016-04-27
Typos

pirapira:
Some small obvious fixes.


DAO.sol
```diff
@@ -530,7 +530,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
         // If the curator removed the recipient from the whitelist, close the proposal
         // in order to free the deposit and allow unblocking of voters
-        if (!isRecipientAllowed(p.recipient) && p.open) {
```
CJentzsch: that's not a typo ;-)
It saves gas to check it here, compared to later.

pirapira: It's already checked on line 524.

PR #139, opened 2016-04-30
Liquid Democracy

jbaylina:
# Notes on Liquid Democracy Pull Request

This pull request tries to implement  liquid democracy (Delegated Demomacry) into the Standard slockit/DAO contract.

The main points on this pull requests are:

### setDelegete( _newDelegate)
This method allows any tokenHolder to setup his delegate.
If the tokenHolder votes, the delegate votes does not count.
If the tokenHolder does not vote, then delegate is taken in account.

### defaultDelegate
Initialy all tokenholders have as a defaultDelegate account.
This defaultDelegate account is defined in the DAO cration.
This defaultDelegate account can be an invalid account if you don't want to use this functionality initially.
This defaultDelegate account can be changed by a proposal.

### Vote census
The vote rights for each poposal is the tokenHolder distribution and delegate assignation at the moment of the proposal creation.
When a token is transfered or a delegator is set, it will take effect only for the newer created proposals.

### No blocking accounts.
With this system the blocking accounts mechanism is no longer necessary.


colm:
wow cool. let me look at this later. Check out https://github.com/slockit/DAO/pull/132 which allows someone to opt out of voting.

LefterisJP:
This is a lot of work and would need quite thorough reviewing. Note that by the time this is reviewed and ready for merging the DAO test framework may have had alterations and you will need to rebase on top of them.

That said this is a lot of work for you first PR into the DAO Standard framework. Very nice and very impressive. I really love that you spent time writting tests too and did not just simply edit the DAO code. Awesome!

Some general remarks:

1) Please squash all commits into one since there are many duplicated named commits
2) Do not commit junk files. Remove the `DS_STORE`, `.ods` files and the generates `.js` test files
3) Do not change arguments to other existing tests, like you did in ` tests/scenarios/split/arguments.json`, if you need this argument changed then add a new argument for the tests you need it in.

I will review more thoroughly later when I find more time.

jbaylina:
@LefterisJP Ok to all your comments!

This PR, has also one problem that should be fixed: This implementation allows to vote on a proposal and then split without taking the consequences of what was voted.  

The solution that I can implement is that in the moment that some body splits, they should:

1.- Execute all accepted proposals. 
2.- Remove all the votes that the DTH or his delegate emitted in the ongoing proposals.

If you think it's interesting, I can continue working on it.

CJentzsch:
Seems to be redundant to https://github.com/slockit/DAO/pull/154. Can we close this? @jbaylina

jbaylina:
This PR has another important advantage, and is that the DTH does not get blocked any more.  

This PR solves the penalty that DTHs get for voting.

I believe that this PR should be at least considered for DAO 2.0.

CJentzsch:
Closed in favour of https://github.com/slockit/DAO/pull/154#issuecomment-224236843

PR #142, opened 2016-05-04
when fired by the client, pay back remaining ether to rewardAccount

CJentzsch:
since the DAOs owning reward tokens own part of that.
But on the other side, if someone split before this proposal got accepted, still owning some rewardTokens from previous projects, he might get quite a reward here which he didn't deserve. 
But the other way around, if someone did split after the proposal got accepted, he should also get part of that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ tests/multisplitrewards.js
 tests/spendall.js
 tests/extrabalance.js
 tests/stealextrabalance.js
+tests/firecontractor.js
 
 # Ignore deployment related file
 deploy/code.js
--- a/SampleOffer.sol
+++ b/SampleOffer.sol
@@ -90,7 +90,7 @@ contract SampleOffer {
 
     // "fire the contractor"
     function returnRemainingMoney() onlyClient {
-        if (client.receiveEther.value(this.balance)())
+        if (client.DAOrewardAccount().call.value(this.balance)())
             promiseValid = false;
     }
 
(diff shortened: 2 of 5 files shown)


LefterisJP:
I agree with this change. This is the most flexible way of handlng the refund when firing.

The DAO can then decide to either claim the refund money for itself as capital or pay it out directly to its members as rewards:
```js
    function retrieveDAOReward(bool _toMembers) external noEther returns (bool _success) {
        DAO dao = DAO(msg.sender);

        if ((rewardToken[msg.sender] * DAOrewardAccount.accumulatedInput()) /
            totalRewardToken < DAOpaidOut[msg.sender])
            throw;

        uint reward =
            (rewardToken[msg.sender] * DAOrewardAccount.accumulatedInput()) /
            totalRewardToken - DAOpaidOut[msg.sender];
        if(_toMembers) {
            if (!DAOrewardAccount.payOut(dao.rewardAccount(), reward))
                throw;
            }
        else {
            if (!DAOrewardAccount.payOut(dao, reward))
                throw;
        }
        DAOpaidOut[msg.sender] += reward;
        return true;
    }
```

LefterisJP:
First things first, rebased on top of latest master to get the newest test changes.

CJentzsch:
rebased

LefterisJP:
I already did that and pushed, I also have rights to the repo :)

LefterisJP:
So the proposal test now works with the changes you made. I also added a much needed test for firing the contractor.

@CJentzsch: Note that due to your rebasing the branch again after I rebased it I had to force-push to update it. If you want to get the changes now you would need to reclone the branch. Or better yet, just merge this PR to master since it's finished and let's continue tests on the Offer in subsequent PRs.

PR #146, opened 2016-05-11
Adds functions implemented on the Ethereum Wallet

alexvandesande:


Token.sol
```diff
@@ -34,6 +34,12 @@ contract TokenInterface {
     mapping (address => uint256) balances;
     mapping (address => mapping (address => uint256)) allowed;
 
+    /// Public variables of the token, all used for display
```
CJentzsch: none of those variables are ever set, so they should be part of the constructor.

alexvandesande: Yes, I don't get this either. Where is the constructor of your functions? Initial supply is not set on this page either

CJentzsch: This is the constructor: https://github.com/slockit/DAO/blob/master/DAO.sol#L354
The DAO contract derives from TokenCreation ( where the "initial supply" gets created) and Token.sol

alexvandesande: Ok, added them

jbaylina:
This variables should also be added also in the DAO_creator.

We should also have to think what should be the token name and token symbols for the split DAOs.

I see two ways:

* Define them in the split proposal.
* Calculate them by adding some suffix: TheDAO_1, TheDAO_2 TheDAO_2_1 for a grandchild and for the symbols: Ð_1,  Ð_2,  Ð_2_1

It would be good also that those names could be changed dynamically by the own DAO.

alexvandesande:
I think if the dao should define these on the split proposal but also be
able to change it later. On our current implementation we only check for
the name when the token is added, we wouldnt auto update when it changes,
but it could be changed.

On Tue, May 24, 2016 at 5:56 AM, Jordi Baylina <notifications@github.com>
wrote:

> This variables should also be added also in the DAO_creator.
>
> We should also have to think what should be the token name and token
> symbols for the split DAOs.
>
> I see two ways:
>
>    - Define them in the split proposal.
>    - Calculate them by adding some suffix: TheDAO_1, TheDAO_2 TheDAO_2_1
>    for a grandchild and for the symbols: Ð_1, Ð_2, Ð_2_1
>
> It would be good also that those names could be changed dynamically by the
> own DAO.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/slockit/DAO/pull/146#issuecomment-221207188>
>



-- 

Alex Van de Sande
UX Designer

jbaylina:
It might be good implement the approveAndCall method like in HumanStandardToken.sol :
https://github.com/ConsenSys/Tokens/blob/master/Token_Contracts/contracts/HumanStandardToken.sol#L50

alexvandesande:
@jbaylina it's there.

LefterisJP:
@alexvandesande 

Thanks a lot for your contribution! But please after rebasing also try to compile the contracts to make sure nothing breaks.  Had to fix some things here: https://github.com/slockit/DAO/commit/a98cc76cc58f5914581094422ccbed091ba59c2e

I need to make the tests, or at least a subset of them running on Travis to avoid situations like this in the future :)

alexvandesande:
Thanks for accepting it Lefteris, sorry for the trouble

On Fri, Jun 3, 2016 at 5:19 AM, Lefteris Karapetsas <
notifications@github.com> wrote:

> @alexvandesande <https://github.com/alexvandesande>
>
> Thanks a lot for your contribution! But please after rebasing also try to
> compile the contracts to make sure nothing breaks. Had to fix some things
> here: a98cc76
> <https://github.com/slockit/DAO/commit/a98cc76cc58f5914581094422ccbed091ba59c2e>
>
> I need to make the tests, or at least a subset of them running on Travis
> to avoid situations like this in the future :)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/slockit/DAO/pull/146#issuecomment-223519197>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAG5Ai5bXyJn3s1a0vT8Ab6cT-A1OqHsks5qH-OVgaJpZM4IcTuh>
> .
>



-- 

Alex Van de Sande
UX Designer

PR #147, opened 2016-05-12
SampleOffer getDailyPayment() + first day fix

LefterisJP:
- Made some fixes to the tests, where the SampleOffer contract in both
  the deployment and the proposal tests was interacting with the
  curator. Now it only interacts with the contractos as it should.

- Fix in tests float with non-float comparison 

- Made a test for SampleOffer that tests the process up to the contractor using getDailyPayment() 
  
- Fix in getDailyPayment() to also alow payment to be withdrawn for the first day.


LefterisJP:
@CJentzsch I am going ahead and merging this. 

If you later want to review it, the only change this PR does to SampleOffer.sol  is this commit: https://github.com/slockit/DAO/pull/147/commits/6c1ca04fa3858005c815db682cdc036452716321

PR #152, opened 2016-05-18
prevent (curator) halving min quorum during fuelling & before first p…

Gustav-Simonsson:
This prevents the DAO curator from halving min quorum during the fuelling and before two weeks after first proposal is submitted, which could be seen as a vulnerability or simply giving the curator more power than I think was initially intended.

In `halveMinQuorum`, the added check `now >= closingTime` is redundant as `proposals.length > 1` enforces this due to the validations in `newProposal` - however figured it would make the code easier to read and audit.


LefterisJP:
Very nice find @Gustav-Simonsson! Thankfully this is not an issue in the currently deployed DAO since neither the curators are malicious, nor do we have much time left for the creation phase.

Let me add a test for this to your PR.

@CJentzsch: Do not merge before I add the tests please.

LefterisJP:
@CJentzsch Tested it, and test is now part of the PR. Merge when you are happy with it. It's already Part of the DAO v2.0 milestone.

waxmihs2902:
awesome

PR #155, opened 2016-05-19
DAOreward payment fixed

jbaylina:
Imagine the next situation:

Initial stete:
 * Main DAO have 90 RewardTokens
 * A splited DAO has 10 RewardTokens.

STEP 1: 100E are sent to the DAORewardToken.Main and SLIPT DAO collect it.

So the state is:
* DAOrewardAccount.accumulatedInput() = 100
* DAOpaidOut[main] = 90
* DAOpaidOut[SlitDAO] = 10
* DAOrewardAccount.balance = 0

STEP 2: A new investment is done of 100E 

 * Main DAO have 190 RewardTokens
 * A splited DAO has 10 Reward Tokens.

STEP 3: 10E Is added to the DAO rewardAccount

* DAOrewardAccount.accumulatedInput() = 110
* DAOrewardAccount.balance = 10

If now the main DAO want's to get his rerard he can't because:

        uint reward =
            (rewardToken[msg.sender] * DAOrewardAccount.accumulatedInput()) /
            totalRewardToken - DAOpaidOut[msg.sender];

reward = (190 *110) / 200 - 90 =  14.5E

and 14.5 < 10 so the transaction will fail.

This PR fix that the main DAO can retrieve this amount.

#### Workarround

1. Wait unitl more Ether is sended to the DAORewardToken
2. Send Ether from the main DAO to the DAORewardAccount in order to later retrieve it (Two proposals).

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -697,7 +697,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
             (rewardToken[msg.sender] * DAOrewardAccount.accumulatedInput()) /
             totalRewardToken - DAOpaidOut[msg.sender];
 
-        if ( DAOrewardAccount.balance < reward) reward = DAOrewardAccount.balance;
+        reward = DAOrewardAccount.balance < reward ? DAOrewardAccount.balance : reward;
 
         if(_toMembers) {
             if (!DAOrewardAccount.payOut(dao.rewardAccount(), reward))
@@ -723,7 +723,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
         uint reward =
             (balanceOf(_account) * rewardAccount.accumulatedInput()) / totalSupply - paidOut[_account];
 
-        if ( rewardAccount.balance < reward) reward = rewardAccount.balance;
+        reward = rewardAccount.balance < reward ? rewardAccount.balance : reward;
 
         if (!rewardAccount.payOut(_account, reward))
             throw;
--- a/DAO.sol
+++ b/DAO.sol
@@ -696,6 +696,9 @@ contract DAO is DAOInterface, Token, TokenCreation {
         uint reward =
             (rewardToken[msg.sender] * DAOrewardAccount.accumulatedInput()) /
             totalRewardToken - DAOpaidOut[msg.sender];
+
+        if ( DAOrewardAccount.balance < reward) reward = DAOrewardAccount.balance;
+
         if(_toMembers) {
             if (!DAOrewardAccount.payOut(dao.rewardAccount(), reward))
                 throw;
@@ -719,6 +722,9 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
         uint reward =
             (balanceOf(_account) * rewardAccount.accumulatedInput()) / totalSupply - paidOut[_account];
+
+        if ( rewardAccount.balance < reward) reward = rewardAccount.balance;
+
         if (!rewardAccount.payOut(_account, reward))
             throw;
         paidOut[_account] += reward;


CJentzsch:
The accumulated input is only increasing, not decreasing when payOuts. are done.
So step3 `DAOrewardAccount.accumulatedInput() = 10` is wrong.

LefterisJP:
@CJentzsch I think he has a typo only in the step. In the calculation he does calculate it as `110`

`reward = (190 *110) / 200 - 90 = 14.5E`

But @jbaylina if you are referring to this check:

```js
        if ((rewardToken[msg.sender] * DAOrewardAccount.accumulatedInput()) /
            totalRewardToken < DAOpaidOut[msg.sender])
            throw;
```
Then the `DAOpaidOut()` would be 90.

And `14.5 < 90` == `false` so there is no `throw`.

Are you satisfied with my analysis @jbaylina? If so can we close the issue?

jbaylina:
Yes, it was a tipo.

LefterisJP:
@jbaylina I understand your concern much better now!

The situation you bumped on is essentially due to the way rewards were originally designed. The DAO rewards account needs to have enough money to distribute to all token holders (no matter if they are in a split DAO child) that funded it in order for the system to be fair. Only then can DAOs call `retrieveDAOReward()`.

So in your scenario the DAOs would only be able to retrieve their rewards after another `100` ETH has been added to the DAORewardAccount.

*STEP 4*: another 100 ETH is added to the reward account
 - `DAOrewardAccount.accumulatedInput()` = 210
 - `DAOrewardAccount.balance` = 110

Only now can any DAO start claiming rewards.

So:

 - parentDAO gets: `(190 * 210) / 200 - 90 ` = `109.5` out of the `110`
 - childDAO gets: `(10 * 210) / 200 - 10` = `0.5` out of the `110`

Why is it made like this? Well imagine that your PR is merged and indeed I can always claim rewards even if the `rewardsAccount` has less than the rewards that my DAO deserves.

In such a scenario a parent DAO, or whichever DAO has the most reward tokens, could starve all other DAOs out of their rewards by repeatedly calling `retrieveDAOReward()` and leaving the account empty.

jbaylina:
@LefterisJP : The main problem with this system is that when you create reward accounts for the main DAO, you are decreasing the percentage of the investment ownership of the child DAOs. If those DAOs already collected the money, they now stay in a negative balance. They should return to the DAOrewardAccount some ether. But of course this will not happen, so the main DAO does not have all the Ether that it should have.
This fix just try to get the remaining Ether in the DAOrewardAccount.

LefterisJP:
@jbaylina You should consider this as one of the disadvantages of splitting.

Let me explain further why I don't think that merging this PR makes sense. Let's assume that this PR is merged and that we are adding a new STEP 3 instead of the one you presented in your introduction.

**STEP 3**
another 95 ETH is added to the reward account

jbaylina:
* DAOrewardAccount.accumulatedInput() = 195
* DAOrewardAccount.balance = 95
* RT mainDAO: 190 
* RT childDAO: 10
* DAOpaidOut[main] = 90
* DAOpaidOut[SlitDAO] = 10

So 

main DAO can get: 

190 * 195 / 200 - 90 = 95,25 (With the patch 95) (without the paths 0)

split DA0 can get

10 *195 / 200 - 10 = -0.25  (With the path and without the path will get 0)

So I see it better with the path.

LefterisJP:
@jbaylina: Yes but please read my answer more carefully. There is one more step where the problem is. Do you see it?

LefterisJP:
@jbaylina: I had a miscalculation when I calculated a bit more complicated scenario which lead me to think that with the patch the parentDAO could starve the childDAO. But I believe this is not the case.

So we are good. Let me make some code-specific comment.

DAO.sol
```diff
@@ -696,6 +696,9 @@ contract DAO is DAOInterface, Token, TokenCreation {
         uint reward =
             (rewardToken[msg.sender] * DAOrewardAccount.accumulatedInput()) /
             totalRewardToken - DAOpaidOut[msg.sender];
+
+        if ( DAOrewardAccount.balance < reward) reward = DAOrewardAccount.balance;
```
LefterisJP: A note on our styling: The expression of the if branch should go to a new line.

But an even better suggestion here is to use the ternary operator:

`reward = DAOrewardAccount.balance < reward ? DAOrewardAccount.balance : reward;`

DAO.sol
```diff
@@ -719,6 +722,9 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
         uint reward =
             (balanceOf(_account) * rewardAccount.accumulatedInput()) / totalSupply - paidOut[_account];
+
+        if ( rewardAccount.balance < reward) reward = rewardAccount.balance;
```
LefterisJP: same here:  `reward = DAOrewardAccount.balance < reward ? DAOrewardAccount.balance : reward;`

LefterisJP:
@CJentzsch What do you think of this PR? Could it be a DAO v2.0 feature?

PR #156, opened 2016-05-19
SampleOffer: Prevent unauthorized value transfer

LefterisJP:
- Add a fallback function to assert that none can send money to
  SampleOffer unless it's done during the `sign()`

- Add a test to assert that simply sending ether to the SampleOffer
  won't work

- Add a `noEther` modifier to all method of `SampleOffer` that should
  not transfer any Ether, to make sure the client can't send any extra
  money via any of them.

- Fix the proposal test by sending the exact amount.

### DIFF
--- a/SampleOffer.sol
+++ b/SampleOffer.sol
@@ -106,6 +106,10 @@ contract SampleOffer {
         client = _newClient;
     }
 
+    function () {
+        throw; // this is a business contract, no donations
+    }
+
     // interface for Ethereum Computer
     function payOneTimeReward() returns(bool) {
         if (msg.value < deploymentReward)
--- a/SampleOffer.sol
+++ b/SampleOffer.sol
@@ -47,6 +47,9 @@ contract SampleOffer {
         _
     }
 
+    // Prevents methods from perfoming any value transfer
+    modifier noEther() {if (msg.value > 0) throw; _}
+
     function SampleOffer(
         address _contractor,
         address _client,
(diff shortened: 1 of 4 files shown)


SampleOffer.sol
```diff
@@ -94,18 +97,22 @@ contract SampleOffer {
...
+    function setDeploymentReward(uint _deploymentReward) onlyClient noEther {
         deploymentReward = _deploymentReward;
     }
 
-    function updateClientAddress(DAO _newClient) onlyClient {
+    function updateClientAddress(DAO _newClient) onlyClient noEther {
         client = _newClient;
     }
 
+    function () {
```
CJentzsch: I agree, no donations, but even in a business contract it should be possible for the client to put more ether into it, for whatever reason (dramatic trop in ether price, extend scope, ...)

CJentzsch:
I agree with everything except of the fallback function.

LefterisJP:
@CJentzsch In that case the contract should be renegotiated and a new Offer deployed in the network. Sending more ether to the same contract should not be possible.

Another reason to not make it possible to send more ether as value transfer is that as @colm pointed out in a conversation we had, some parties within the DAO could make small proposals to transfer a little ether to the offer, after it has been signed in order to spread FUD and discredit the contractor by making up god knows what kind of rumours.

As a general rule, I believe that if you can imagine it happening, it will probaly happen so it's better to take measures against it.

And for any renegotiation I think that a new contract should be deployed.

PR #158, opened 2016-05-22
Refactor modifiers

colm:
We should centralize this checking. I believe the contract was half written before modifiers were implemented. So its a nice way to clean up a little.


LefterisJP:
I like the idea, but only for modifiers that are used frequently and would make the code easier to read.

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+    }
+    modifier Fueled{
+          if(!isFueled) throw;
+            _
+    }
+    modifier notDAO {
+        if (msg.sender == address(this)) throw;
+            _
+    }
+    modifier onlyDAO {
```
LefterisJP: This `notDAO` for example is used only in one place where the is also a comment that explains why the check is there. By using a modifier we gain nothing and lose a lot on readability.

So I'd rather not have this as a modifier.

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+        if (msg.sender != address(this)) throw;
+            _
+    }
+    modifier whiteList(address _newContract) {
```
LefterisJP: Also I think you should remove this too. Because:
1) It's used in only one place as a modifier.
2) Modifiers that take arguments make the code hard to read, in my opinion

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
     }
+    modifier creationOver{
+          if (now < closingTime) throw;
+	    _
+    }
+    modifier Fueled{
+          if(!isFueled) throw;
+            _
+    }
+    modifier notDAO {
```
LefterisJP: This `notDAO` for example is used only in one place where the is also a comment that explains why the check is there. By using a modifier we gain nothing and lose a lot on readability.

So I'd rather not have this as a modifier.

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
         if (balanceOf(msg.sender) == 0) throw;
             _
     }
+    modifier creationOver{
+          if (now < closingTime) throw;
+	    _
+    }
+    modifier Fueled{
```
LefterisJP: let's have consistency, make this lowercase `fueled`

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
         if (balanceOf(msg.sender) == 0) throw;
             _
     }
+    modifier creationOver{
```
LefterisJP: Maybe `afterCreation` is a better name here?

DAO.sol
LefterisJP: Also space between `fueled` and `{`

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+          if (now < closingTime) throw;
```
LefterisJP: styling: `throw` should be in a new line under the `if`

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+	    _
+    }
+    modifier Fueled{
+          if(!isFueled) throw;
```
LefterisJP: styling: `throw` should be in a new line under the `if`

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+            _
+    }
+    modifier notDAO {
+        if (msg.sender == address(this)) throw;
```
LefterisJP: styling: `throw` should be in a new line under the `if`

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+            _
+    }
+    modifier onlyDAO {
+        if (msg.sender != address(this)) throw;
```
LefterisJP: styling: `throw` should be in a new line under the `if`

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+            _
+    }
+    modifier whiteList(address _newContract) {
+        if (!allowedRecipients[_newContract]) throw;
```
LefterisJP: styling: `throw` should be in a new line under the `if`

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+            _
+    }
+    modifier notBlocked {
+        if (isBlocked(msg.sender)) throw;
```
LefterisJP: styling: `throw` should be in a new line under the `if`

DAO.sol
```diff
@@ -350,6 +349,34 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+            _
+    }
+    modifier onlyCurator {
+        if (msg.sender != curator) throw;
```
LefterisJP: styling: `throw` should be in a new line under the `if`

colm:
Thanks for the feedback. Made those changes.

LefterisJP:
Thanks! But wow, 7 commits for this change is a bit too much. Can you squash all these fixes in one: "PR review changes" commit?

CJentzsch:
Do you think it increases readability?
If so, please rebase, if not, please close ;-)

PR #161, opened 2016-05-23
pay reward to original DAO

CJentzsch:
rewards are always paidOut to original DAO, to ensure that the new DAO doesn't rob the splitDAOs of there rewards

### DIFF
--- a/SampleOffer.sol
+++ b/SampleOffer.sol
@@ -36,6 +36,7 @@ contract SampleOffer {
 
     uint public dateOfSignature;
     DAO public client; // address of DAO
+    DAO public originalClient; // address of DAO who signed the contract
     bool public isContractValid;
 
     uint public rewardDivisor;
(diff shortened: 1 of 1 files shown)


LefterisJP:
@CJentzsch Good point but I got a question here.

What if the `originalClientDAO` wants to change to a totally different DAO that does not come out of a split and call `updateClientAddress()` for that DAO?

Then all rewards will be paid back to the originalDAO but how can the new client claim them?

CJentzsch:
Thats exactly the reason I made this for. SplitDAO *as well as* the original DAO with an updated contract  hold reward tokens and can get the rewards from the mother DAO.

LefterisJP:
I know, I was thiking more of a totally different DAO, that does not hold any reward tokens since it did not come from `split` or `update`. 

But that is probably not a good idea to do and we should advise against it.

LefterisJP:
@CJentzsch I added a comment to `updateClientAddress()` to make this even clearer.

This PR is a must-have and as far as I am concerned has the :+1: to be merged.

PR #162, opened 2016-05-24
Propagate reward tokens to descendent DAOs

jbaylina:
This PR tries to solve the problem with the Reward Tokens propagation in the split of a split DAO.

This implementation also allows reward tokens owners of a DAO to transfer his reward tokens to a different contract or to a regular account.

If this PR is merged, some tests should be done.

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -665,8 +665,15 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
 
         // Assign reward rights to new DAO
-
         this.transferRewardTokens(p.splitData[0].newDAO, balances[msg.sender], p.splitData[0].totalSupply);
+
+        // Transfer also reward tokens of all ancestors.
+        address daoAddress = privateCreation;
+        while (daoAddress != 0) {
+            DAO ancestorDAO = DAO(daoAddress);
+            ancestorDAO.transferRewardTokens(p.splitData[0].newDAO, balances[msg.sender], p.splitData[0].totalSupply);
+            daoAddress = ancestorDAO.privateCreation();
+        }
 
         // Burn DAO Tokens
         Transfer(msg.sender, 0, balances[msg.sender]);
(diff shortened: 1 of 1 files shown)


DAO.sol
```diff
@@ -715,6 +717,45 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+        while (daoAddress != 0) {
+            DAO parentDAO = DAO(daoAddress);
+            parentDAO.retrieveDAOReward(_toMembers);
+            daoAddress = parentDAO.privateCreation();
+        }
+
+        return true;
+    }
+
+    function transferRewardTokens(address _to, uint _amountNum, uint _amountDen) external noEther returns (bool _success) {
```
CJentzsch: with this you make rewardTokens freely tradeable, and DAO tokens value becomes very hard to distinguish.

DAO.sol
```diff
@@ -715,6 +717,45 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+
+        if (   (_amountDen == 0)
+            || (_amountNum > _amountDen))
+            throw;
+        uint rewardTokensToTransfer = rewardToken[msg.sender] * _amountNum / _amountDen;
+
+        rewardToken[msg.sender] -= rewardTokensToTransfer;
```
CJentzsch: this needs to be called by the motherDAO, why should they do it? Theres is no incentivization for it.

CJentzsch:
Not convinced that this solves the problem yet. Please convince me ;-)
If not, lets close it.

jbaylina:
transferRewardTokens is a recursive function. It transfer RT to a new address. But not only the RT of the self DAO, also the parent RTs owned by self, the grandparent RTs owned by self and so on. You can see that this function is called befor split to transfer the RT of self and all ancestors to the child DAO.

Reward tokens might be tradable now also. One way to do it is: If you set the curator and the DAO tokens owner of a child DAO to a single walled contract instead of an account. Then you just transfer/trade the ownership of the wallet.

If you want to limit, I don't understand exactly why, you can always limit this call to another contract.

collectInheritedDAOrewards is called by the child to get the rewards of the parent, grandparent and so on DAOs. This is the easy way for the child DAOs to collect all the rewards of all ancestors with a single call.

jbaylina:
I simplified and fixed a bug on this patch.  transferRewardTokens is not recursive any more and just transfers the reward tokens of this dao belonging to msg.self to another account

In the split process you transfer the reward tokens of the self DAO and all the reward tokens of all the ancestors that belongs to self.

collectInheritedDAOrewards is called by a child DAO to get the rewards of the parent, grandparent and all the ancestors DAOs. This is the easy way for the child DAOs to collect all the rewards of all ancestors with a single call.

CJentzsch:
cleaning up PRs (real solution to fair reward per project distribution is more complex)

PR #169, opened 2016-05-24
Fix `getDailyPayment()` calculation

LefterisJP:
- If we are at the end of the payment period and we can withdraw more
  than the balance, simply take the balance and not have the money stuck
  there forever.


waxmihs2902:
nice

PR #170, opened 2016-05-25
Improve Offer withdraw mechanics

chfast:


SampleOfferWithoutReward.sol
```diff
@@ -101,6 +136,14 @@ contract SampleOfferWithoutReward {
             paidOut += amount;
     }
 
+    function withdraw() {
+        var daysSinceSignoff = now - dateOfSignature; // TODO: Can be negative?
```
LefterisJP: `daysSinceSignOff` needs a division by 1 days here, since now it's seconds since sign off

LefterisJP:
@chfast General comments:

1. Adding comments is one PR on its own.
2. Remove the TODOs, if anyone wants to copy the contract and change the names he is free to do so.
3. You spotted the bug #171, if you would like provide a fix suggestion in a sepate PR

So in short this PR can be either for adding comments, or for fixing the bug you found, not both :)

SampleOfferWithoutReward.sol
```diff
@@ -90,8 +124,9 @@ contract SampleOfferWithoutReward {
             isContractValid = false;
     }
 
+    // TODO: This function is quite inconvenient. See withdraw() alternative.
     function getDailyPayment() {
-        if (msg.sender != contractor)
```
LefterisJP: This is needed here mostly for semantics.

Indeed the money will end up with the contractor no matter who ends up calling this function, but the intention of this contract is that only the contractor should call this. Thus we have checks for anything that should not happen, like someone else calling the function for the contractor to withdraw money from the contract.

SampleOfferWithoutReward.sol
```diff
@@ -101,6 +136,14 @@ contract SampleOfferWithoutReward {
             paidOut += amount;
     }
 
+    function withdraw() {
```
LefterisJP: I don't understand why try to make a different function for this. Put your improvements, if any, inside `getDailyPayment()`

SampleOfferWithoutReward.sol
chfast: If someone wants to sponsor a transaction for the contractor, why not let him?

LefterisJP: Because this is a contract between the contractor and the client and nobody else.

Even if it's not a threat to the security of the contract, I would like to have checks for anything that is not supposed to happen.

SampleOfferWithoutReward.sol
```diff
@@ -101,6 +136,14 @@ contract SampleOfferWithoutReward {
...
+        var daysSinceSignoff = now - dateOfSignature; // TODO: Can be negative?
+        // FIXME: The amount ignores hirtorical changes to dailyWithdrawLimit.
+        var amount = (oneTimeCosts + daysSinceSignoff * dailyWithdrawLimit) - paidOut;
```
CJentzsch: `onetimeCosts` are not included in `paidOut` . Therefore the amount is way too high, Essentially paying out the oneTimeCosts twice.

CJentzsch:
fixed by https://github.com/slockit/DAO/pull/173

Issue #171, opened 2016-05-25
Withdraw of negative amount from SampleOffer


chfast:
You can withdraw negative amount of money from `SampleOffer.getDailyPayment()` because the calculations ignores historical changes to `dailyWithdrawLimit`.

chfast:
@CJentzsch I can send a fix.

LefterisJP:
Awesome find Pawel :)

PR #173, opened 2016-05-25
SampleOffer: fix withdraw negative amount issue

chfast:
This fixes the issue https://github.com/slockit/DAO/issues/171. Instread of remembering how much was payed out, it is remembered when the last payout was. Type type of withdraw limits has been also changed to uint128 to avoid overflow calculating the amount for the withdraw.

### DIFF
--- a/SampleOfferWithoutReward.sol
+++ b/SampleOfferWithoutReward.sol
@@ -48,12 +48,12 @@ contract SampleOfferWithoutReward {
 
     // The minimal daily withdraw limit that the Contractor accepts.
     // Set once by the Offerer.
-    uint public minDailyWithdrawLimit;
+    uint128 public minDailyWithdrawLimit;
 
     // The amount of money the Contractor has right to withdraw daily above the
     // initial withdraw. The Contractor does not have to do the withdraws every
     // day as this amount accumulates.
-    uint public dailyWithdrawLimit;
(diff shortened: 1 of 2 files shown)


SampleOfferWithoutReward.sol
```diff
@@ -118,15 +124,21 @@ contract SampleOfferWithoutReward {
...
+    // Withdraw to the Contractor.
+    //
+    // Withdraw the amount of money the Contractor has right to according to
+    // the current withdraw limit.
+    // Executing this function before the Offer is signed off by the Client
+    // makes no sense as this contract has no money.
+    function getDailyPayment() noEther {
+        uint timeSinceLastPayment = now - lastPayment;
+        // Calculate the amount using 1 second precision.
+        uint amount = (timeSinceLastPayment * dailyWithdrawLimit) / (1 days);
```
CJentzsch: you are calculating "seconds since last payment" time `dailyWithdrawLimit` . This is way too much.
It neds to be : `(timeSinceLastPayment  / (1 days)) * dailyWithdrawLimit; `

LefterisJP:
@chfast Thank you for finding the bug and for providing a nice fix. Now time to test.

CJentzsch:
@chfast Thanks a lot!

PR #174, opened 2016-05-25
avoid negative withdrawal

CJentzsch:


CJentzsch:
fixed by https://github.com/slockit/DAO/pull/173

PR #175, opened 2016-05-25
Prevent unauthorized calls of getDailyPayment()

LefterisJP:


SampleOfferWithoutReward.sol
```diff
@@ -131,6 +131,10 @@ contract SampleOfferWithoutReward {
     // Executing this function before the Offer is signed off by the Client
     // makes no sense as this contract has no money.
     function getDailyPayment() noEther {
+        if (msg.sender != contractor && msg.sender != address(client)) {
```
CJentzsch: is the message coming from the client or the SampleOffer contract itself?

LefterisJP: Very good question, the code I wrote assumes client. Let's check that out.

chfast: It is good idea.

However, let's consider removing the `getDailyPayment()` call from the `setDailyWithdrawLimit()`. We should not have issues like before (please verify that!), only the Contractor get less money for the period from the last payout. But:
- the DAO can consider that,
- the Contractor will have the time to payout because the voting will take some time.

chfast: I changed my mind about checking if the sender is the contractor. This checks gives as proof that the contractor still has its private key and no money goes to void. And this is nice property.

LefterisJP: Exactly! Good point.

But still I would like to verify our understanding of who is `msg.sender` in such a case as Christoph pointed.

For the moment we are testing the changes introduced by your Previous PR.

chfast: I guess the `msg.sender` is not changes because solidity implements internal calls without the `CALL` instruction.

LefterisJP: Yes that seems to be the case.

chfast: Anyway, that's my alternative.

waxmihs2902:
thanks

SampleOfferWithoutReward.sol
chfast: https://github.com/slockit/DAO/pull/177

CJentzsch:
fixed by https://github.com/slockit/DAO/pull/177

PR #177, opened 2016-05-25
SampleOffer: decouple withdraw and changing withdraw limit

chfast:
Not performing the withdraw to the Contractor before changing the withdraw limit by the Client is very minor issue. If the Contractor has not done a withdraw recently:
- the Client can take that into account and advice the Contractor to do it,
- the decision about changing the withdraw limit is done by voting so will take some time - during this period the Contractor has time to do a withdraw by the current limit.

### DIFF
--- a/SampleOfferWithoutReward.sol
+++ b/SampleOfferWithoutReward.sol
@@ -109,13 +109,8 @@ contract SampleOfferWithoutReward {
     }
 
     function setDailyWithdrawLimit(uint128 _dailyWithdrawLimit) onlyClient noEther {
-        if (_dailyWithdrawLimit >= minDailyWithdrawLimit) {
-            // Before changing the limit withdraw the money the Contractor has
-            // right to. The payment may not be accepted by the Contractor but
-            // it is the Contractor's problem.
-            getDailyPayment();
+        if (_dailyWithdrawLimit >= minDailyWithdrawLimit)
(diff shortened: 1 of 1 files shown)


LefterisJP:
Thanks, this is cleaner and we also keep the semantics of who can withdraw payment from the contract, clean.

So essentially the contractor has to perform the withdraw before the proposal debate period of `setDailyWithdrawLimit()` is done (2 weeks). This should be giving him ample time to do so.

But what is gained by decoupling it apart from cleaner code? The way I see this is we get cleaner/less code and we add a complication:

1. In the case the contractor withdraws right before the limit change all is fine.

2. In the case he has not withdrawn for 15 days and let's say the client is trying to make a:
    i. Proposal to *lower* the limit: Then the contractor will end up with less money than he is entitled for the period of those 15 days.
    ii. Proposal to *increase* the limit: Then the contractor will end up with more money than he is entitled to for the period of those 15 days.

Both `2i` and `2ii` are not really serious problems, but why introduce them with this PR?

chfast:
This PR just reverts some changes done in https://github.com/slockit/DAO/pull/173. So we have the same semantics as before the fix. But the bug was fixed.

Solving 2 would need more thinking.

LefterisJP:
@chfast Didn't calling `getDailyPayment()` from inside `setDailyWithdrawLimit()` effectively solve 2?

PR #178, opened 2016-05-25
SampleOffer: Remove unneeded variable

LefterisJP:
@CJentzsch: The DAO can't set any divisor fee anymore, right? So this variable is a forgotten remnant that should be removed, correct?


CJentzsch:
its very needed. Fix: https://github.com/slockit/DAO/pull/180

PR #179, opened 2016-05-25
Sample offer quorum for firing

LefterisJP:
### **Warning** In progress. Do not merge.

 This PR is currently hampered by [this](https://github.com/ethereum/solidity/issues/598) solidity compiler bug and will not work, until it's fixed.
---
### The problem

Firing the contractor, or changing the client can currently be done with the DAO's minimum quorum which is only 20% of all the voters. That could be considered inconvenient, even though any such proposals can also be downvoted.
### The solution

At the deployment of the offer, the proposal also has an argument `_quorumForChange` which is a number from 0 to 100 representing the percentage of quorum needed for the execution of the important functions.

Both `returnRemainingEther()` and `updateClientAddress()` will also take the `_proposalID` of the proposal that voted them in as an argument. Then the Offer contract can query the client DAO and see if the proposal got passed with a quorum as big as the required `quorumForChange`. If and only if that is the case they will be executed.

To properly achieve this, the `proposalHash`, the `yea` and the `nay` of the proposal have to be checked.
### Potential downsides of the approach

Passing the `_proposalID` as an argument to the proposal may be considered inconvenient. If someone else tries to put in a proposal to the DAO at the same block as you, you may end up with a different ID than you expected, and have to remake the proposal.

But I believe that the ability to impose specific quorum for contract-breaking changes such as changing the client or firing the contractor, is a benefit that far outweighs the aforementioned inconvenience.
---

**NOT READY YET - DO NOT MERGE**

### DIFF
--- a/SampleOffer.sol
+++ b/SampleOffer.sol
@@ -36,14 +36,16 @@ contract SampleOffer is SampleOfferWithoutReward {
             bytes32 _IPFSHashOfTheProposalDocument,
             uint _totalCosts,
             uint _oneTimeCosts,
-            uint128 _minDailyWithdrawLimit
+            uint128 _minDailyWithdrawLimit,
+            uint _quorumForChange
         ) SampleOfferWithoutReward(
             _contractor,
             _client,
             _IPFSHashOfTheProposalDocument,
             _totalCosts,
             _oneTimeCosts,
-            _minDailyWithdrawLimit) {
+            _minDailyWithdrawLimit,
+            _quorumForChange) {
         }
 
     // interface for Ethereum Computer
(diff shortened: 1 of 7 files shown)


colorics:
Help me ! I did - Proposal_ID: 9 (DAO split proposed)
TimeStamp: 8 hrs 9 mins ago (2016-05-29 12:59:52 +00:00)
Created By: 0x228d29ea776cb17ca0db0538562ecaacdc0a9e46
Recipient: 0x228d29ea776cb17ca0db0538562ecaacdc0a9e46
This purse -https://etherscan.io/address/0xc0f5228127fe170c6c26e46a1462a86403fd726c - vote - YES!
He wants to steal my - ETH -? What should I do???

CJentzsch:
@LefterisJP can we close this?

LefterisJP:
@CJentzsch No I may rebase this in the future, just leave it as is for now.

CJentzsch:
clean up

PR #183, opened 2016-05-26
Forbid original client to create reward tokens

CJentzsch:


LefterisJP:
This is needed if the original DAO moves to a new contract and all its reward tokens are zero. It has my vote.

:+1:

PR #184, opened 2016-05-26
Separate USN interface to different contract to avoid rewardToken attack

CJentzsch:


USN_RewardPayOut.sol
```diff
@@ -0,0 +1,31 @@
+import "./SampleOffer.sol";
+
+contract USN_RewardPayOut {
```
LefterisJP: Remove snakecase. Make it `USNRewardPayOut` all over the contract.

USN_RewardPayOut.sol
```diff
@@ -0,0 +1,31 @@
...
+     
+     SampleOffer usnContract;
```
LefterisJP: make that: `SampleOffer public usnContract`;

SampleOffer.sol
```diff
@@ -53,35 +53,4 @@ contract SampleOffer is SampleOfferWithoutReward {
     function setDeploymentReward(uint _deploymentReward) onlyClient noEther {
```
colm: Maybe we can put this back into `SampleOfferWithoutReward`

LefterisJP: No that would not really make sense since that's for a contract that should have no rewards, and this is the rewards extension.

colm: The diff is just so small.

LefterisJP:
I performed a rebase on master to incorporate the removal of the getters from the offer contracts.

LefterisJP:
Do not merge until I have also added tests for this. It is gonna take some time, since it will require a few changes.

LefterisJP:
Tests added and passing. It should be okay now, you can merge.

SampleOffer.sol
```diff
@@ -55,43 +55,11 @@ contract SampleOffer is SampleOfferWithoutReward {
...
-
     function getRewardDivisor() noEther constant returns (uint) {
         return rewardDivisor;
     }
 
     function getDeploymentReward() noEther constant returns (uint) {
         return deploymentReward;
     }
-
-    // interface for USN
```
chfast: If this is moved is the offer with the reward still needed?

USNRewardPayOut.sol
```diff
@@ -0,0 +1,32 @@
...
+          usnContract = _usnContract;
+     }
+
+     // interface for USN
+    function payOneTimeReward() returns(bool) {
+        if (msg.value < usnContract.getDeploymentReward())
+            throw;
+
+          if (usnContract.getOriginalClient().DAOrewardAccount().call.value(msg.value)()) {
+            return true;
```
chfast: If the only return value is `true` returning it makes no sense.

USNRewardPayOut.sol
```diff
@@ -0,0 +1,32 @@
...
+contract USNRewardPayOut {
+     
+     SampleOffer public usnContract;
+     
+     function USNRewardPayOut(SampleOffer _usnContract) {
+          usnContract = _usnContract;
+     }
+
+     // interface for USN
+    function payOneTimeReward() returns(bool) {
```
chfast: The `payReward()` function can do the same (and more). Why do you need both?

LefterisJP: It has to do with our proposal to the Universal Sharing Network. This contract ... the `USNRewardPayout` won't be deployed at the moment, it will just use the deployed offer so it may change later.

The difference is that the nodes of the USN , the EC(Ethereum Computers) can either pay an one time Fee to the DAO or just pay a percentage.

We will work on this further in the future, and keep it a separate contract to give us the flexibility to do so, plus to avoid the "DAO using the offer to generate endless reward tokens" problem.

chfast: Shouldn't you remember that a EC paid one time reward then?

USNRewardPayOut.sol
```diff
@@ -0,0 +1,32 @@
...
+        if (msg.value < usnContract.getDeploymentReward())
```
chfast: If the deployment reward is strictly related to the USN, why not having it in this contract?

USNRewardPayOut.sol
LefterisJP: That remembering is going to be in yet another contract ;)

USNRewardPayOut.sol
LefterisJP: Because:

1) The DAO needs to be able to vote and set those at any value it seems fit, so it needs to be in the offer contract which has to be in the whitelist. Keep in mind it is just a `SampleOffer` and you can and should modify/add/remove parameters for your own proposal as you see fit.

2) This contract `USNRewardPayOut` is separate and will not be in the whitelist and needs no integration with the DAO.

PR #188, opened 2016-05-27
Add split grace + quorum change

D-Nice:
Features and changes requested for V1.1 of the DAO

- added a grace period, to allow for splitting from the DAO, if an unfavorable proposal passes. The splitting grace period constant should be greater than the minimum debate/vote time for split, their difference is the amount of time someone has, to set up a split before funds go to the proposal. With the current constants (10 days, 1 week respectively) you have 3 days, after a proposal's voting ends, to make a minimum debate period split proposal, which allows you to split, before the proposal can get executed.
- changed only yea's to be counted towards quorum (may need to lower minquorum, although it can be halved which should suffice).

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -29,6 +29,8 @@ contract DAOInterface {
     // The amount of days for which people who try to participate in the
     // creation by calling the fallback function will still get their ether back
     uint constant creationGracePeriod = 40 days;
+    // Period of time contract execution is locked to allow splitting, make greater than minSplitDebatePeriod
+    uint constant splittingGracePeriod = 10 days;
     // The minimum debate period that a generic proposal can have
     uint constant minProposalDebatePeriod = 2 weeks;
     // The minimum debate period that a split proposal can have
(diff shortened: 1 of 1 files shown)


DAO.sol
```diff
@@ -516,7 +521,12 @@ contract DAO is DAOInterface, Token, TokenCreation {
         uint waitPeriod = p.newCurator
             ? splitExecutionPeriod
             : executeProposalPeriod;
-        // If we are over deadline and waiting period, assert proposal is closed
+                
+        // Add grace period to wait time if proposal appears passable
+        if (p.yea > p.nay)
+           waitPeriod += splittingGracePeriod;
```
colm: Is it possible for an attacker to call `executeProposal` repeatedly. Then each time `waitPeriod` will be increased?

D-Nice: No it shouldn't, because waitPeriod is reinitialized at every executeProposal call at line 521.

D-Nice: An extra check could be added to the if statement at line 526 to avoid any possible attacks or vulnerabilities of waitPeriod being summed over and over again, although I don't see the possibility of such an attack- 
waitPeriod == executeProposalPeriod

D-Nice:
Time to distance oneself from this.

PR #189, opened 2016-05-27
Offer: remove redundant valid flag

chfast:


LefterisJP:
Even though this change would not hurt I would like to avoid any non-essential changes to the contract at this time.

@CJentzsch: ?

SampleOfferWithoutReward.sol
```diff
@@ -160,8 +158,7 @@ contract SampleOfferWithoutReward {
 
     // "fire the contractor"
     function returnRemainingEther() onlyClient {
-        if (originalClient.DAOrewardAccount().call.value(this.balance)())
-            isContractValid = false;
+        originalClient.DAOrewardAccount().call.value(this.balance)();
```
colm: You need to wrap send functions in if statments because of the call stack attack http://martin.swende.se/blog/Devcon1-and-contract-security.html

chfast: 1. This attack cannot harm you here. It's all transactional, the transfer to the DAO reward account is done or not.
2. Solidity is not affected by this attack.

chfast: @LefterisJP @CJentzsch @colm confirm / deny?

chfast:
I don't understand why this flag is in public API. Does it provides any useful information? You have never defined what it actually means. Is the state where the offer has some ethers and is marked as invalid with this flag possible?

When the _avoid any non-essential changes_ period ends?

chfast:
Rebased.

chfast:
Rebased. Again. Any opinions about this?

Issue #160, opened 2016-05-22
Feature Request: make DAO.sol modular based upon permission


colm:
All the variables and functions that individual users can change should be defined in a separate contract for `transfer*` `newProposal` `vote` `executeProposal` `splitDAO` `retrieveDAOReward` `getMyReward` `withdrawRewardFor` `halveMinQuorum`

`DAO.sol` should define the functions and variables that a Majority of the Users can changes. Changing proposal deposit `changeProposalDeposit`, `newContract`

'Curators.sol' should define the functions and variables that the Curator can change. `changeAllowedRecipients` `halveMinQuorum`

This is a good idea because
1. It will make reviewing a little simpler as you can easily confirm that all the functions in `Curator.sol` are only accessible to the curators. Also the files will be smaller and considering the modules separately can make things less complex. 

2. There are many use cases where only changes to UserState.sol (EG:(Identity Corp)[https://blog.ethereum.org/2013/12/31/bootstrapping-a-decentralized-autonomous-corporation-part-3-identity-corp/] ) would be required. 

3. Although this reduces the readability a little, it make the understanding the framework and its implications easier. If forces the reader to consider that it is basically a state machine, some variables can be changed by individuals, some by the majority (if you don't like a change you can split to a new DAO which is initialized with a certain state based upon 'splitDAO') this is a very powerful paradigm. These changes will allow developers to more easily leverage the DAO framework.

jbaylina:
An intermediate state could be just sort them and put a Header Comment before each block.

Before splitting more the code in different files, I would do some kind of generateDeployScript script in order  to make the deployment of the different contracts more easy. (geth , mist). This is specially annoying when you are manually testing.

PR #192, opened 2016-05-28
moved contracts from DAO.sol

colm:
First step to https://github.com/slockit/DAO/issues/160


LefterisJP:
Squash your commits please. This is just a code movement and should not be split over 2 commits.

Other comment: I see where you are coming from with this change, but splitting it into even more files will make verification by the solidity web browser even more tedious. More files to copy from into one combined contract.

dirtyfilthy:
Tedious? Surely combining files is just a case of combineFiles.sh

CJentzsch:
I need to redo this completely because a lot has changed

PR #195, opened 2016-05-30
returnRemainingEther() should not allow value transfer

LefterisJP:
- Add a check in `returnRemainingether()` for disallowing value transfr

- Add a test for this scenario

### DIFF
--- a/SampleOfferWithoutReward.sol
+++ b/SampleOfferWithoutReward.sol
@@ -159,7 +159,7 @@ contract SampleOfferWithoutReward {
     }
 
     // "fire the contractor"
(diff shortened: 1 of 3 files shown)


colm:
I did not think that this was such a problem as the next line returns `this.balance` although it is possible that someone could use a call stack attack to make this fail. Its probably better to have `noEther` nice catch.

LefterisJP:
No this is a serious problem. Returning `this.balance` includes the extra value you send with `returnRemainingEther()`. That means that you would create `msg.value` free reward tokens.

colm:
yes you are right.

PR #197, opened 2016-05-30
prevent tokens from being sent to the dao itself

p-s-dev:


CJentzsch:
why?

D-Nice:
It dilutes the overall value of tokens, and they won't contribute to votes.

p-s-dev:
1. People send on accident, thinking that is how to vote
2. mist doesn't sum tokens correctly when dao holds them (granted, that is a mist "bug")

p-s-dev:
Currently, people have accidentally sent 56,828 dao tokens (worth ~$7,330) to TheDao.

DAO.sol
```diff
@@ -740,6 +740,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
         if (isFueled
             && now > closingTime
             && !isBlocked(msg.sender)
+            && _to != address(this)
```
CJentzsch: you would allso need that change in `transferFrom`

p-s-dev: Done, added to transferFrom() also

PR #198, opened 2016-05-31
Allow public creation after split

p-s-dev:


p-s-dev:
need to squash commits before making PR

PR #200, opened 2016-05-31
Stop potential recursive executeProposal() call

LefterisJP:
Such a scenario would be very easily spotted and stopped
during the security vetting of the recipient contract and such a
contract would never make it in the whitelist.

Nevertheless, in the case of an obviously malicious contract that calls
executeProposal() recursively making into the whitelist, we are moving
the `proposalPassed` guard variable before the call() to the recipient
contract.

This will protect against the recipient contract calling
`executeProposal()` again recursively and effectively emptying the DAO
from funds.

Thanks for spotting this @chriseth!

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -526,6 +526,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
         if (now < p.votingDeadline  // has the voting deadline arrived?
             // Have the votes been counted?
             || !p.open
+            || p.proposalPassed // anyone trying to call us recursively?
             // Does the transaction code match the proposal?
             || p.proposalHash != sha3(p.recipient, p.amount, _transactionData)) {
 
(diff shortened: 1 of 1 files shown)


D-Nice:
Yet if the send call were to throw, the funds from the proposal could never be sent out and the proposal becomes invalid.

LefterisJP:
@D-Nice Sure, but the problem this PR solves is in the case it does not throw. Check the description.

D-Nice:
Isn't there a more elegant solution that doesn't break current features, such as ensuring the executeProposal function isn't called more than once within a certain amount of blocks? Wouldn't that stop any recursion exploits?

D-Nice:
@LefterisJP The recursive attack would happen within the same block, right? In that case, could you simply not make a counter that ensures the executeProposal function is never called more than once, for a specific proposal within a block?

LefterisJP:
@D-Nice What feature do you think is broken here?

D-Nice:
@LefterisJP It renders the following feature useless:  
    // (used in the case `executeProposal` fails because it throws)
    uint constant executeProposalPeriod = 10 days;

Basically the executeProposal wait period, which before this change, would allow someone to call executeProposal again, in case even the send call failed. Now if the send call fails 
    (if (!p.recipient.call.value(p.amount)(_transactionData)))
the proposal becomes permanently invalid, as it will throw an exception for any future attempted calls. In this case, you may as well just close the proposal, as it has the same effect, once p.proposalPassed = true, it's practically the same thing, but less elegant.

LefterisJP:
@D-Nice I don't think you understand how the EVM works in this particular instance.

The affected function is not a `send`, it's a `call`. Specifically here:
```
            if (!p.recipient.call.value(p.amount)(_transactionData))
                throw;
```

If a transaction throws, all calls get reversed. Everything gets reversed. Even what happened right before it. So `p.proposalPassed` will still be `false`.

And you also misunderstand the purpose behind the `executeProposalPeriod`. It's essentially a time after which you can close the proposal by calling `executeProposal()`.

```
        // If we are over deadline and waiting period, assert proposal is closed
        if (p.open && now > p.votingDeadline + waitPeriod) {
            closeProposal(_proposalID);
            return;
        }
```

Issue #203, opened 2016-05-31
It is difficult to discover what code will be executed if a proposal has passed


dirtyfilthy:
Currently on proposal creation, a proposalHash is created via sha3(_recipient, _amount, _transactionData) to enforce that the right code is called during a call to executeProposal(). While this might save space, it also makes it difficult when browsing proposals to see what _transactionData consisted of without examining the original transaction that created the proposal. 

_transactionData should be added as a field to the Proposal struct, or at the very least, the data should be propagated by the ProposalAdded() event, so that various automated tools in use by the community can make it available (dao.report et al)

CJentzsch:
solved in https://github.com/slockit/DAO/pull/228

Issue #80, opened 2016-04-10
 `executeProposal` takes `_transactionData` as input not `p._transactionData`


colm:
Because `executeProposal` takes `_transactionData` as a function input and not from `p._transactionData` [here](https://github.com/slockit/DAO/blob/master/DAO.sol#L488) and [here](https://github.com/slockit/DAO/blob/master/DAO.sol#L509)

There is no problem here. However because [sha3](http://solidity.readthedocs.org/en/latest/units-and-global-variables.html#mathematical-and-cryptographic-functions) concatenates without padding an attacker can inject a malicious `_transactionData` into the `executeProposal` function along as it is concentrated with `p.recipient` and `p.amount` the check will catch it. But I would be a little more comfortable if we used `p._transactionData` here.

Is there a reason we take `_transactionData` as input?

LefterisJP:
Is this issue just for a question? Because as you pointed out [this check](https://github.com/slockit/DAO/blob/master/DAO.sol#L488) will catch any malicious `_transactionData` injection.

But yes I am not sure why we `_transactionData` is given as an argument here and not saved inside the proposal itself as `p.transactionData`. @CJentzsch can help us with that.

colm:
Just a question. That check catches it. I would just rather it it was `p.transactionData`

CJentzsch:
There is no `p.transactionData`. I want to avoid typos, so when executing the one doing it makes sure he does so on the right proposal. The voters can always use `checkProposalCode` to check whether they vote on the right proposal. Does this answer your question @colm ?

LefterisJP:
@CJentzsch But isn't it better to have it as `p.transactionData`? And then the only thing the user who executes the proposal has to do is check the proposalD is correct? 

This way we can remove:
1) The `bytes _transactionData` argument from `executeProposal()`
2) The `p.proposalHash` member will no longer be needed
3) The check `|| p.proposalHash != sha3(p.recipient, p.amount, _transactionData)) {`

CJentzsch:
Sure, but we have this mechanism exactly because of the reason to avoid accidently executing the wrong proposal.this was build from Christian and Avsa in the original DAO contract.

LefterisJP:
So it's a choice between having an extra safeguard (apart from checking that you are executing the correct proposal ID) and reducing the complexity of the code by removing the 3 bullet points I showcased [above](https://github.com/slockit/DAO/issues/80#issuecomment-208270241).

[...]

PR #205, opened 2016-06-02
Add transaction data to Proposal and ProposalAdded Event

dirtyfilthy:
This commit solves issue #203. Currently it is difficult to see what code will be executed if a  proposal is passed. Adding transactionData to the Proposal struct and the ProposalAdded event solves this problem, and allows third party tools to display this data easily. Adding transactionData to the Proposal struct was  previously discussed  in the comments of issue #80. There seems to be no good reason why the transactionData is not contained with the Proposal struct.

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -112,8 +112,6 @@ contract DAOInterface {
         // True if quorum has been reached, the votes have been counted, and
         // the majority said yes
         bool proposalPassed;
-        // A hash to check validity of a proposal
-        bytes32 proposalHash;
         // Deposit in wei the creator added when submitting their proposal. It
         // is taken from the msg.value of a newProposal call.
         uint proposalDeposit;
@@ -131,6 +129,9 @@ contract DAOInterface {
         mapping (address => bool) votedNo;
         // Address of the shareholder who created the proposal
         address creator;
+        // transactionData to be passed as part of call to recipient upon
+        // proposal execution
+        bytes transactionData;
     }
 
     // Used only in the case of a newCurator proposal.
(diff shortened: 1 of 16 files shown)


LefterisJP:
Hm ... yes I don't see a reason for not having this in the future. As we discussed in #80 the arguments for keeping the `proposalHash` instead of `transactionData` were mostly weak, and as of now since the contract is already deployed, irrelevant. (we don't fear to make code changes as long as they are not too big, and easily testable)

But a few points in your PR.

1. Your indentation is off. Use the same indentation as the rest of the codebase, which is multiples of 4 spaces.

2. You should also remove: `proposalHash` from the code totally. That means remove it from the struct, remove all places where it is used instantiated and also remove `checkProposalCode()` since it would make no more sense as a function.

3. In `executeProposal()` you can now also remove `bytes _transactionData`. If this is now a member of the struct the function can just read it from the proposal struct itself.

@CJentzsch: How do you feel with this making it in the next version of the code that is deployed?
I personally think it's totally fine, plus as we have witnessed from the tools that are being developed it would make life easier for the developers of said tools.

DAO.sol
```diff
@@ -440,6 +444,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
         p.recipient = _recipient;
         p.amount = _amount;
         p.description = _description;
+		p.transactionData = _transactionData;
```
CJentzsch: only spaces, no tabs please

dirtyfilthy:
Thanks for the feedback, i have implemented your suggestions and also fixed up the tests.

LefterisJP:
Thank you for your work! You will also need to rebase on top of the latest master since there have been some changes.

dirtyfilthy:
done.

LefterisJP:
Thank you. You may need to do it again if the contract diverges too much. The changes in the tests should not be done at the moment. I am going to work on preparing the testing framework to be able to both:

1) Test against the deployed DAO v1.0 contract
2) Test against the latest master branch DAO contract.

To achieve this I need to do quite a bit of work, so it may take some time till I can accept your PR. But it's in the pipeline and will do it as soon as I can.

@CJentzsch Do not merge until we have the ability to test both things I wrote above.

LefterisJP:
Please rebase on latest develop. Also please remove any changes you made to the tests. They are not needed and will break the tests at the moment. Let me handle them since the required changes will be a bit more complicated.

CJentzsch:
(needs to be rebased and reopened)

PR #207, opened 2016-06-02
only allow oneTimeCosts to be paid once

CJentzsch:


PFOffer.sol
```diff
@@ -45,6 +45,7 @@ contract PFOffer {
     // signed.
     // Set once by the Offerer.
     uint oneTimeCosts;
+    bool oneTimeCostsPayed;
```
LefterisJP: Here, and in the rest of the code:
`Payed` -> `Paid`

PFOffer.sol
```diff
@@ -183,11 +184,12 @@ contract PFOffer {
     }
 
     function getOneTimePayment() noEther {
-        if (msg.sender != contractor || now < dateOfSignature + 3 weeks )
+        if (msg.sender != contractor || now < dateOfSignature + 3 weeks || oneTimeCostsPayed )
             throw;
 
         if (!contractor.send(oneTimeCosts))
             throw;
+        oneTimeCostsPayed = true;
```
LefterisJP: Even though it's just a `send` and not a `call`, so probably gas stipend would not be enough to do anything bad maybe move `oneTimeCostsPaid = true;` before the `if (!contractor.send(oneTimeCosts))` ?

Just like we did here: https://github.com/slockit/DAO/pull/200

PR #209, opened 2016-06-02
Use isBlocked() to stop addresses from receiving Ð

GriffGreen:
This is to prevent users from locking up tokens that they want to vote with accidentally when they transfer them to their wallets. it will forsce them to Create a new wallet...

This is a suggestion and should be thought about further.... It fixes this issue, preventing users from sending tokens to addresses that are locked, but it also requires them to practice good key and password management (backing up their new private keys and remembering their passwords) this is not a no brainer.


LefterisJP:
Yep I like the idea, we can leave the PR open, no hurry anyway, and I should add a test for it too in the PR.

GriffGreen:
:-) I feel so cool adding features like I know what I'm doing ;-)

PR #211, opened 2016-06-02
remoevd call.value

colm:
removed call.value to arbitrary address.


p-s-dev:
msg.sender.call.value(weiGiven[msg.sender])()
should change to 
msg.sender.send(weiGiven[msg.sender])
not 
msg.sender.send(weiGiven[msg.sender])()
right?  the extra "()" at the end gets removed for send()

colm:
Good Catch. Updated. thanks for chekcing ;-)

PR #217, opened 2016-06-06
Add checkVoteStatus to PFOffer.sol

LefterisJP:
This PR is made to add the functionality of what @alexvandesande introduced in [his sample Offer contract](https://gist.github.com/alexvandesande/122c05a0f8a7108523aed29ae7ec92b1) as `protectFromAmbush()` to the Proposal Framework Offer contract located in this repository.

The name of the function is changed to `checkVoteStatus()`. At any point within `voteStatusDeadline` seconds before the end of the voting period anyone can call `checkVoteStatus()` to attempt and set the
`wasApprovedBeforeDeadline` flag.

This allows for a way to protect  against a last minute yes vote  by requiring the flag to be set to true at an earlier point of the voting deadline. The offer can not be signed without this flag being set.

Tests for the functionality and logic introduced by `checkVoteStatus()` are added and are passing.

**a note for the compilation of the contract**: You need a very recent version of solc. It needs to contain [this](https://github.com/ethereum/solidity/pull/624) fix for the [Dynamic Return Data Decoder](https://github.com/ethereum/solidity/issues/598#issuecomment-224015639) bug.

### DIFF
--- a/PFOffer.sol
+++ b/PFOffer.sol
@@ -217,8 +217,6 @@ contract PFOffer {
 
     // The proposal will not accept the results of the vote if it wasn't able
     // to be sure that YEA was able to succeed 48 hours before the deadline
-    function checkVoteStatus() {
-        var (,,votingDeadline,,,,,,yea,nay,) = client.proposals(proposalID);
     function checkVoteStatus() noEther {
         var (,,,votingDeadline,,,,,,yea,nay,) = client.proposals(proposalID);
         uint quorum = (yea + nay) * 100 / client.totalSupply();
--- a/PFOffer.sol
+++ b/PFOffer.sol
@@ -38,6 +38,9 @@ contract PFOffer {
 
     // Period of time after which money can be withdrawn from this contract
     uint constant payoutFreezePeriod = 3 weeks;
+    // Time before the end of the voting period after which
+    // checkVoteStatus() can no longer be called
+    uint constant voteStatusDeadline = 48 hours;
 
     // The total cost of the Offer. Exactly this amount is transfered from the
     // Client to the Offer contract when the Offer is signed by the Client.
@@ -72,6 +75,8 @@ contract PFOffer {
     DAO client; // address of DAO
     DAO originalClient; // address of DAO who signed the contract
     bool isContractValid;
+    uint proposalID;
+    bool wasApprovedBeforeDeadline;
 
     modifier onlyClient {
         if (msg.sender != address(client))
@@ -88,7 +93,8 @@ contract PFOffer {
         bytes32 _hashOfTheProposalDocument,
         uint _totalCosts,
         uint _oneTimeCosts,
-        uint128 _minDailyWithdrawLimit
+        uint128 _minDailyWithdrawLimit,
+        uint _proposalID
     ) {
         contractor = _contractor;
         originalClient = DAO(_client);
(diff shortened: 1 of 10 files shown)


PFOffer.sol
```diff
@@ -98,6 +118,7 @@ contract PFOffer {
         oneTimeCosts = _oneTimeCosts;
         minDailyWithdrawLimit = _minDailyWithdrawLimit;
         dailyWithdrawLimit = _minDailyWithdrawLimit;
+        proposalID = _proposalID;
```
alexvandesande: The contractors doesn't know the proposal id at this point at it might take a while until it's whitelisted. They should have a separate functon `watchContract` or similar that will allow it.

LefterisJP: Yes I have considered that. It's easy to get the proposal ID. It's just always `proposalsNumber()+1`.

The problem with this approach is that you gotta make the proposal right after you deploy the contract. And even when you do you may collide with someone else who is making a proposal at the same time.


I had not thought of a separate `watchContract` function. That's a nice idea. But we need to have a way to prevent the contractor from setting it to a proposalID that's not theirs in order to fool the `checkVoteStatus()` function.

Probably the easiest way to do so is by means of community awareness. If the proposalID does not match the one that actually tries to sign() the contract we can put out a community alert that the contractor is a scammer.

alexvandesande: Check how I implemented the watch function to prevent that:

https://gist.github.com/alexvandesande/122c05a0f8a7108523aed29ae7ec92b1#file-proposal-js-L271

It checks if the address is the same and if the proposal has not been voted
yet. Someone could still trick it by submitting two proposals, but then the
token holders would be able to see that.

On Tue, Jun 7, 2016 at 9:02 AM, Lefteris Karapetsas <
notifications@github.com> wrote:

> In PFOffer.sol
> <https://github.com/slockit/DAO/pull/217#discussion_r66057366>:
>
> > @@ -98,6 +118,7 @@ contract PFOffer {
> >          oneTimeCosts = _oneTimeCosts;
> >          minDailyWithdrawLimit = _minDailyWithdrawLimit;
> >          dailyWithdrawLimit = _minDailyWithdrawLimit;
> > +        proposalID = _proposalID;
>
> Yes I have considered that. It's easy to get the proposal ID. It's just
> always proposalsNumber()+1.
>
> The problem with this approach is that you gotta make the proposal right
> after you deploy the contract. And even when you do you may collide with
> someone else who is making a proposal at the same time.
>
> I had not thought of a separate watchContract function. That's a nice
> idea. But we need to have a way to prevent the contractor from setting it
> to a proposalID that's not theirs in order to fool the checkVoteStatus()
> function.
>
> Probably the easiest way to do so is by means of community awareness. If
> the proposalID does not match the one that actually tries to sign() the
> contract we can put out a community alert that the contractor is a scammer.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/slockit/DAO/pull/217/files/5620e943d777294e89ca7e1263df220847a10a92#r66057366>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAG5AtCRRr7jH2XcVM5fm5YwwEVdi_LYks5qJV3rgaJpZM4IvUqK>
> .
>



-- 
Alex Van de Sande
Lead Designer
Ethereum Foundation

LefterisJP: Yes I had seen your implementation, but as you also point out it can be tricked by submitting two different proposals.

At the end I think that all these will be so easily detectable by the token holders that it does not matter at all. If such a thing happens the DAO observation team will put out an alert.

I am gonna port the function as is :)

PR #226, opened 2016-06-08
Image added to the README_DTHPool

jbaylina:


LefterisJP:
An image is worth a thousand words :+1:

PR #228, opened 2016-06-08
Dao11

CJentzsch:
Big PR :-)

I will make many small PRs for every feature implemented here soon.

I implemented the following features:

1) withdraw (soloSplit)
2) Grace period after voting deadline
3) 2 days preSupport verification
4) DHT can remove addresses from the whitelist
5) withdraw means remove vote
6) ability to take back / change a vote (except of the 2 days freeze)
7) only count yes votes in quorum
8) Remove extraBalance
9) add dependencies in proposal

A wiki explaining them is work in progress


DAO.sol
```diff
@@ -121,6 +130,12 @@ contract DAOInterface {
         bool newCurator;
         // Data needed for splitting the DAO
         SplitData[] splitData;
+        // dependencies in other proposals. List of proposal IDs and list defining whether the other proposal needs to have passed or failed
```
LefterisJP: Split comment in 2 lines

DAO.sol
```diff
@@ -121,6 +130,12 @@ contract DAOInterface {
...
+        uint[] dependencies;
+        bool[] dependenciesType;
+        // true if more tokens are in favour of the proposal than opposed to it at
+        // least 2 days before the voting deadline
```
LefterisJP: replace 2 days with `preSupportTime`

DAO.sol
```diff
@@ -121,6 +128,13 @@ contract DAOInterface {
         bool newCurator;
         // Data needed for splitting the DAO
         SplitData[] splitData;
+        // Dependencies in other proposals. List of proposal IDs and list defining
+        // whether the other proposal needs to have passed or failed
+        uint[] dependencies;
```
LefterisJP: Why not do:

``` js
struct Dependency {
    uint proposalID;
    bool shouldPasss;
}
```

and then have it as:

``` js
Depency[] dependencies
```

inside the proposal struct?

If you insist on keeping it as is then change:
1. `dependencies` -> `dependentProposalIDs`
2. `dependenciesType` -> `dependencyShouldPass`

Because as it currently stands the name are confusing.

DAO.sol
```diff
@@ -511,9 +535,56 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+        }
+
+        votingRegister[msg.sender].length = 0;
+        blocked[msg.sender] = 0;
+    }
+
+    function verifyPreSupport(uint _proposalID) {
+        Proposal p = proposals[_proposalID];
+        if (now < p.votingDeadline - preSupportTime) {
+            if (p.yea > p.nay) {
```
LefterisJP, line 580: Maybe add a `yea` quorum requirement at the presupport? Just like we have done with `PFOffer.sol`?

DAO.sol
```diff
@@ -566,18 +645,20 @@ contract DAO is DAOInterface, Token, TokenCreation {
                 proposalCheck = false;
         }
 
-        if (quorum >= minQuorum(p.amount)) {
+        uint minimalQuorum = minQuorum(p.amount);
```
LefterisJP: here rename `minimalQuorum` to `minimumQuorum`

DAO.sol
```diff
@@ -566,18 +645,20 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+
+        if (quorum >= minimalQuorum) {
```
LefterisJP: here rename `minimalQuorum` to `minimumQuorum`

DAO.sol
```diff
@@ -566,18 +645,20 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
-            if (quorum > totalSupply / 5)
-                minQuorumDivisor = 5;
+            // set the minQuorum to 14.3% again, in the case it has been reached
+            if (quorum > totalSupply / 7)
+                minQuorumDivisor = 7;
         }
 
         // Execute result
-        if (quorum >= minQuorum(p.amount) && p.yea > p.nay && proposalCheck) {
+        if (quorum >= minimalQuorum && p.yea > p.nay && proposalCheck) {
```
LefterisJP: here rename `minimalQuorum` to `minimumQuorum`

DAO.sol
CJentzsch, line 580: I think it is enough that we don't count the nay votes in the real quorum. 
Are there any good arguments for a yea quorum here?

LefterisJP, line 580: Same one as I used for PFOffer.sol. I can call `verifyPreSupport()` in the beginning when the vote is not really set and set the flag to `true` if yea > nay.

At a later state someone else may call it to change it back to false if yea and nay are in a very close fight. But if they don't, then the `preSupport` flag will be set while it should not have been.

CJentzsch, line 580: As you said, the function can be called again and again to update the result, until `preSupportTime` before the end of the voting deadline. I think this is enough, another quorum is not needed in my view.

DAO.sol
```diff
@@ -826,32 +948,23 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
 
     function changeAllowedRecipients(address _recipient, bool _allowed) noEther external returns (bool _success) {
-        if (msg.sender != curator)
+        if (msg.sender != curator && (msg.sender != address(this) || _allowed))
```
LefterisJP, line 951: This desperateley needs a comment.

So now the DAO can call `changeAllowedRecipient()` only to remove someone from the whitelist?

DAO.sol
```diff
@@ -586,7 +590,6 @@ contract DAO is DAOInterface, Token, TokenCreation {
             // related addresses. Proxy addresses should be forbidden by the curator.
             if (p.recipient != address(this) && p.recipient != address(rewardAccount)
                 && p.recipient != address(DAOrewardAccount)
-                && p.recipient != address(extraBalance)
                 && p.recipient != address(curator)) {
```
kangaroosterus, line 677: Do you think this should be removed: `&& p.recipient != address(extraBalance)`? If we remove it I think we will be double counting reward tokens.

I think what we want to do is move the funds from the extraBalance account to the DAO without incrementing the reward tokens initially. Then when there is a real proposal that requires funding from the DAO, then the reward tokens will be incremented appropriately.

kangaroosterus, line 677: In addition to my comment above. How do we actually move funds from the extraBalance to the DAO?

If a Proposal's recipient is the DAO itself, then the code in executeProposal() will send funds to itself when executing:

```
   if (!p.recipient.call.value(p.amount)(_transactionData))
      throw;
```

So it seems like we need some special code where the DAO can send a message to the extraBalance account by calling `extraBalance.payout(address(this), extraBalance.balance)`

DAO.sol
```diff
@@ -607,52 +610,40 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
         uint _proposalID,
         address _newCurator
     ) noEther onlyTokenholders returns (bool _success) {
-		
-		// Did you already vote on another proposal?
+
+        unblockMe();
+
+        // Did you already vote on another proposal?
         if (blocked[msg.sender] != _proposalID && blocked[msg.sender] != 0)  {
```
kangaroosterus: Not sure what this check does `'blocked[msg.sender] != _proposalID`. Can the caller simply find out what proposalID is blocking them and call withdraw() with that proposal id which would allow them to bypass this check?

This same check is also made in splitDao() and you may want to see if it make sense there as well.

DAO.sol
```diff
@@ -607,52 +610,40 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
-    function soloSplit(
+
+    function withdraw(
         uint _proposalID,
         address _newCurator
     ) noEther onlyTokenholders returns (bool _success) {
-		
-		// Did you already vote on another proposal?
+
+        unblockMe();
```
kangaroosterus: May also want to make this call at the beginning of DAO.splitDAO().

DAO.sol
kangaroosterus: I would also add the following logic near the top of this withdraw() function so that we don't have to execute any of the remaining code in the function:

```
if( balances[msg.sender] <= 0)
    throw;
```

May also want to make the same logic in `splitDAO()`.

LefterisJP: This is checked by the `onlyTokenholders` modifier

LefterisJP: > Not sure what this check does 'blocked[msg.sender] != _proposalID. Can the caller simply find out what proposalID is blocking them and call withdraw() with that proposal id which would allow them to bypass this check?

This `(blocked[msg.sender] != _proposalID && blocked[msg.sender] != 0)` essentially stops the caller from splitting if they are blocked in any proposal. So the check is also in the beginning of splitDao already.

kangaroosterus: >  essentially stops the caller from splitting if they are blocked in any proposal

If that's the goal we only need: `blocked[msg.sender]  !=  0`

But the current logic is `(blocked[msg.sender] != _proposalID && blocked[msg.sender] != 0)` and that logic can be controlled one way or the other by the caller since they can essentially call `withdraw()` with any proposal ID. So for example let's say the msg.sender is blocked on proposal 5. They call withdraw() with proposal id = 5 and they we will not enter the if statement. If they want they can decide to go into that 'if' statement by calling withdraw() with a proposal id of 6.

DAO.sol
kangaroosterus, line 580: I think there's an issue with unvote() if you don't keep state of their original balance[voter] used to increment yea or nay votes.

Here's the scenario:

1) balances[voter] is currently 100
2) the msg.sender votes yes and yea is incremented by 100
3) someone calls Token.transfer(voter, 1000) or the spender.transferFrom(someone, voter, 1000)
4) we unvote() based on a balance[vote] of 1000 and yea is decremented too much.

Perhaps we can use Proposal.yesVote[voter] and Proposal.votedNo[voter] can be used to keep state of their original voting balance[voter] instead of as a bool. 

Thoughts?

DAO.sol
```diff
@@ -615,6 +695,47 @@ contract DAO is DAOInterface, Token, TokenCreation {
         p.open = false;
     }
 
+    function withdraw(
+        uint _proposalID,
```
LefterisJP: @CJentzsch Why do you give `_proposalID` here? It's not used inside the function.

DAO.sol
LefterisJP: @cougarously You are correct.
@CJentzsch The unvote function needs the proposals to keep how much each user voted.
So we need:

```js
        // Simple mapping to check if a shareholder has voted for it
        mapping (address => bool) votedYes;
        // Simple mapping to check if a shareholder has voted against it
        mapping (address => bool) votedNo;
```

to change to:

```js
        // Simple mapping to check if a shareholder has voted for it
        mapping (address => int) votedYesAmount;
        // Simple mapping to check if a shareholder has voted against it
        mapping (address => int) votedNoAmount;
```

in order to be able to keep the original voting state in case someone else transfers tokens to or from the voter's address.

LefterisJP, line 580: @cougarously: No we don't need this because we are now also stopping people from transferring tokens to blocked addresses.

so in your scenario step (3) would fail.

kangaroosterus, line 580: @LefterisJP - Ah okay! I see that there is DAO.transfer and DAO.transferFrom() that overrides the ones in Token.

kangaroosterus, line 580: How about this scenario.

I have a voted on many proposals that are currently still active. One of those proposals is a split proposal and I split and I'm not part of the DAO anymore but I have influence on the proposals that are still active. Should this be allowed? If not (which I believe is the case), we need to unvoteAll()  when burning their tokens in splitDAO() and withdraw().

So here's what may need to be done:

1) add the modifier onlyTokenholders to unvote() and unvoteAll()
2) In the splitDAO() and withdraw(), call unvoteAll() first before setting balance[msg.sender] = 0. However, you want a way to _not_  unvote the split proposal itself so that you can keep a record.

TokenCreation.sol
```diff
@@ -102,13 +97,11 @@ contract TokenCreation is TokenCreationInterface, Token {
...
         if (now < closingTime && msg.value > 0
-            && (privateCreation == 0 || privateCreation == msg.sender)) {
+            && (parentDAO == 0 || parentDAO == msg.sender)) {
 
-            uint token = (msg.value * 20) / divisor();
-            extraBalance.call.value(msg.value - token)();
+            uint token = msg.value;
             balances[_tokenHolder] += token;
             totalSupply += token;
-            weiGiven[_tokenHolder] += msg.value;
```
pirapira, line 111: `weiGiven` should be removed from `TokenCreationInterface` too.

DAO.sol
```diff
@@ -511,9 +535,56 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+
+        if (p.votedNo[msg.sender]) {
+            p.nay -= balances[msg.sender];
+            p.votedNo[msg.sender] = false;
+        }
+    }
+
+    function unVoteAll() {
+        for (uint i = 0; i < votingRegister[msg.sender].length; i++) {
+            Proposal p = proposals[i];
```
pirapira: This line should lookup `votingRegister[msg.sender][i]` somehow.

LefterisJP: @pirapira I think you are right here. [This](https://github.com/slockit/DAO/pull/233) PR should solve it.

DAO.sol
CJentzsch, line 580: Thats already the case. In splitDAO and withdraw `unvoteAll` is called.

DAO.sol
CJentzsch, line 677: In the new DAO there is no extraBalance anymore.

DAO.sol
```diff
@@ -910,6 +1023,23 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
         return isBlocked(msg.sender);
     }
+
+    function combinedBalanceOf(address _address){
+        DAO(parentDAO).balanceOf(_address) + balanceOf(_address);
+    }
+
+    // approve DAO to transfer your tokens prior to that
+    function swapTokens() {
+        uint balance = DAO(parentDAO).balanceOf(msg.sender);
```
colm: I think here `msg.sender` is the address of the current DAO as `swapTokens` is a function call.

DAO.sol
pirapira: Yes it does it.

DAO.sol
```diff
@@ -511,9 +535,56 @@ contract DAO is DAOInterface, Token, TokenCreation {
             blocked[msg.sender] = _proposalID;
         }
 
+        votingRegister[msg.sender].push(_proposalID);
         Voted(_proposalID, _supportsProposal, msg.sender);
     }
 
+    function unVote(uint _proposalID){
```
colm, line 542: can we make `unVote` internal? I think its better to minimize the attack surface in this way. Unless you intend for it to be called by DTH.

LefterisJP, line 542: I think it's intended to be used by the DTH too.

colm, line 542: So then I think `unVote` should remove the proposal ID from `votingRegister[msg.sender]` Otherwise if someone votes on too many proposals it will result in them being unable to call `unVoteAll` due to block gas limit. Then they cannot call `unVote` to remove once blockage at a time since `votingRegister[msg.sender]` remains the same.

DAO.sol
```diff
@@ -826,32 +947,24 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
 
     function changeAllowedRecipients(address _recipient, bool _allowed) noEther external returns (bool _success) {
-        if (msg.sender != curator)
+        // only allow the curator and the DAO itself (only removing addresses) to make changes to the whitelist
+        if (msg.sender != curator && (msg.sender != address(this) || _allowed))
```
colm, line 951: `msg.sender != address(this)` is a pointless check. If `msg.sender == curator` we know `msg.seder != address(this)` maybe it should be `_recipient` to disallow removal of the dao from whitelist.

LefterisJP, line 951: The second leg of the boolean condition after the && is essentially saying:

> And if the DAO makes a proposal to change the allowed recipient and that change is removal from the whitelist then let it go through.

DAO.sol
CJentzsch, line 542: this is done in `unvoteAll` . This can be called at any time to clear things up.

CJentzsch:
closed in favour of:
https://github.com/slockit/DAO/pull/244 https://github.com/slockit/DAO/pull/245 https://github.com/slockit/DAO/pull/246 https://github.com/slockit/DAO/pull/247 https://github.com/slockit/DAO/pull/248 https://github.com/slockit/DAO/pull/249 https://github.com/slockit/DAO/pull/250

PR #230, opened 2016-06-08
added getter-functions for Voting UI

simon-jentzsch:
new functions:
- lastVote()
- numberOfDependencies()
- requiredProposalId()
- requiredProposalType()

changed from internal to constant
- isBlocked()
- minQuorum()

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -159,6 +159,9 @@ contract DAOInterface {
         DAO newDAO;
     }
 
+    // used when returning the last VotingResult
+    enum VoteResult { NotVoted, Yes, No, NoTokens }
+
     // Used to restrict access to certain functions to only DAO Token Holders
     modifier onlyTokenholders {}
 
@@ -244,6 +247,11 @@ contract DAOInterface {
     /// @param _proposalID The proposal ID
     /// @param _supportsProposal Yes/No - support of the proposal
     function vote(uint _proposalID, bool _supportsProposal);
+
+    /// @notice returns the last Vote on proposal `_proposalID` 
+    /// @param _proposalID The proposal ID
+    /// @return _result NotVoted/Yes/No/NoTokens 
+    function lastVote(uint _proposalID) constant returns (VoteResult _result);
 
     /// @notice Checks whether proposal `_proposalID` with transaction data
     /// `_transactionData` has been voted for or rejected, and executes the
@@ -331,6 +339,19 @@ contract DAOInterface {
     /// achieved in 52 weeks
     /// @return Whether the change was successful or not
     function halveMinQuorum() returns (bool _success);
+
+    /// @return total number of dependencies of a proposal
+    function numberOfDependencies(uint _proposalID) constant returns (uint _numberOfDependencies); 
+
+    /// @param _proposalID Id of the proposal
+    /// @param _index the index within the dependencies
+    /// @return the id of the Proposal it depends on
+    function requiredProposalId(uint _proposalID,uint _index) constant returns (uint _requiredProposalID); 
+
+    /// @param _proposalID Id of the proposal
+    /// @param _index the index within the dependencies
+    /// @return the type of the Proposal it depends on
+    function requiredProposalType(uint _proposalID,uint _index) constant returns (bool _requiredProposalType);
 
     /// @return total number of proposals ever created
     function numberOfProposals() constant returns (uint _numberOfProposals);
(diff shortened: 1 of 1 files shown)


DAO.sol
```diff
@@ -341,7 +362,7 @@ contract DAOInterface {
 
     /// @param _account The address of the account which is checked.
     /// @return Whether the account is blocked (not allowed to transfer tokens) or not.
-    function isBlocked(address _account) internal returns (bool);
+    function isBlocked(address _account) constant returns (bool);
```
LefterisJP: This is not a nice change, people should not be able to call isBlocked() as a constant, since it actually changes state.

We need to:
1. rename `isBlocked()` to `queryOrModifyBlocked()` since it can actually change the blocked state:
2. Create a constant `isBlocked()` function

simon-jentzsch: I agree, I just didn't see it changing the state. But it should renamed then!

LefterisJP: Exactly! And you can do it in this PR :)

DAO.sol
```diff
@@ -332,6 +340,19 @@ contract DAOInterface {
     /// @return Whether the change was successful or not
     function halveMinQuorum() returns (bool _success);
 
+    /// @return total number of dependencies of a proposal
+    function numberOfDependencies(uint _proposalID) constant returns (uint _numberOfDependencies); 
+
+    /// @param _proposalID Id of the proposal
+    /// @param _index the index within the dependencies
+    /// @return the id of the Proposal it depends on
+    function requiredProposalId(uint _proposalID,uint _index) constant returns (uint _requiredProposalID);
```
LefterisJP: Capitalize ID.  `requiredProposalId` -> `requiredProposalID`.
Also put space after ','

DAO.sol
```diff
@@ -332,6 +340,19 @@ contract DAOInterface {
...
+
+    /// @param _proposalID Id of the proposal
+    /// @param _index the index within the dependencies
+    /// @return the id of the Proposal it depends on
+    function requiredProposalId(uint _proposalID,uint _index) constant returns (uint _requiredProposalID); 
+
+    /// @param _proposalID Id of the proposal
+    /// @param _index the index within the dependencies
+    /// @return the type of the Proposal it depends on
+    function requiredProposalType(uint _proposalID,uint _index) constant returns (bool _requiredProposalType);
```
LefterisJP: Space after comma here

DAO.sol
```diff
@@ -539,6 +560,15 @@ contract DAO is DAOInterface, Token, TokenCreation {
         Voted(_proposalID, _supportsProposal, msg.sender);
     }
 
+    function lastVote(uint _proposalID) constant returns (VoteResult _result) {
+        Proposal p = proposals[_proposalID];
+        if (p.votedYes[msg.sender])     return VoteResult.Yes;
```
LefterisJP: Love this function! But please put braces and new lines in the if statements <3

DAO.sol
```diff
@@ -999,6 +1028,18 @@ contract DAO is DAOInterface, Token, TokenCreation {
         );
     }
 
+    function numberOfDependencies(uint _proposalID) constant returns (uint _numberOfDependencies) {
+        return proposals[_proposalID].dependencies.length;
+    }
+
+    function requiredProposalId(uint _proposalID,uint _index) constant returns (uint _requiredProposalID) {
```
LefterisJP: put space after `,`

DAO.sol
```diff
@@ -999,6 +1028,18 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+       return proposals[_proposalID].dependencies[_index];
+    }
+
+    function requiredProposalType(uint _proposalID,uint _index) constant returns (bool _requiredProposalType) {
```
LefterisJP: put space after `,`

colm:
What I would really like is that if when a function does not change the state that we preface it with `get` like  `getLastVote`

The reason is for sec reasons I don't ahve to review these functions. I just need to ensure tehy don't change anything. It makes it a little easier to review.

LefterisJP:
@colm I always hated getters which need to have the `get` prefix. If it fits the name then we should have it. If not it's not a nice idea to force it in there. As far as the security perspective is concerned I sure hope that despite the name of the function the actual contents are always read and checked :)

colm:
Yes thats okay. It is just mildly annyoing getting side tracked to confirm what a function does white reading the code. This tends to happen alot for me. And since DAO.sol is getting bigger its more annoying. But its fine the way it is. I would love to move DAOinterface to a seperate file.

PR #231, opened 2016-06-08
allow splits through proposal offer

CJentzsch:


PFOffer.sol
```diff
@@ -182,10 +182,12 @@ contract PFOffer {
     }
 
     function sign() {
+        var (recipient,,,votingDeadline,,) = client.proposals(proposalID);
         if (msg.sender != address(originalClient) // no good samaritans give us ether
             || msg.value != totalCosts    // no under/over payment
             || dateOfSignature != 0       // don't sign twice
-            || !wasApprovedBeforeDeadline)// fail if the voteStatusCheck was not done
+            || !wasApprovedBeforeDeadline // fail if the voteStatusCheck was not done
+			|| now < votingDeadline + 3 days)
```
LefterisJP: Please use spaces not tabs

PFOffer.sol
```diff
@@ -182,10 +182,12 @@ contract PFOffer {
     }
 
     function sign() {
+        var (recipient,,,votingDeadline,,) = client.proposals(proposalID);
```
LefterisJP: `recipient` is not used so you can replace it with a `_`. Don't leave it empty, the contract won't compile then.

PFOffer.sol
CJentzsch: ops

PFOffer.sol
```diff
@@ -253,7 +255,7 @@ contract PFOffer {
     // to be sure that YEA was able to succeed 48 hours before the deadline
     function checkVoteStatus() noEther {
         var (,,,votingDeadline,,,,,,yea,nay,) = client.proposals(proposalID);
-        uint quorum = (yea + nay) * 100 / client.totalSupply();
+        uint quorum = yea * 100 / client.totalSupply();
```
LefterisJP: :+1:

PR #232, opened 2016-06-09
Bugfix: Prevent exploits using reentrant calls from `send`.

chriseth:


LefterisJP:
Thank you Christian :)

PR #237, opened 2016-06-11
Add an optional payout freeze to Offer.sol

LefterisJP:
Tests still need to be adjusted for this change.

### DIFF
--- a/Offer.sol
+++ b/Offer.sol
@@ -54,6 +54,9 @@ contract Offer {
     // withdrawals every day as this amount accumulates.
     uint128 dailyWithdrawalLimit;
 
+    // Period of time after which money can be withdrawn from this contract
+    uint payoutFreezePeriod;
+
     // The address of the Contractor.
     address contractor;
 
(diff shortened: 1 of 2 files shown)


Offer.sol
```diff
@@ -177,16 +195,32 @@ contract Offer {
...
         uint timeSincelastWithdrawal = now - lastWithdrawal;
         // Calculate the amount using 1 second precision.
         uint amount = (timeSincelastWithdrawal * dailyWithdrawalLimit) / (1 days);
         if (amount > this.balance) {
             amount = this.balance;
         }
-        if (contractor.send(amount))
-            lastWithdrawal = now;
+        var lastWithdrawalReset = lastWithdrawal;
+        lastWithdrawal = now;
```
CJentzsch: The 2 lines above can be removed. What if `contractor.send(amount)` fails. then you would have set the wrong value.

LefterisJP: if it fails lastWithdrawalReset would be reset inside the if

PR #238, opened 2016-06-11
add splt grace period to offer.sol

CJentzsch:


Offer.sol
```diff
@@ -173,6 +178,19 @@ contract Offer {
         isContractValid = true;
     }
 
+    // Once a proposal is submitted, the Contractor should call this
+    // function to get the voting deadline of the proposal
+    function getVotingDeadline(uint _proposalID) noEther {
```
colm: should be `setVotingDeadline`

CJentzsch:
fixed by https://github.com/slockit/DAO/pull/239

PR #239, opened 2016-06-11
add splt grace period to offer.sol - alternative

CJentzsch:
alternative to https://github.com/slockit/DAO/pull/238

### DIFF
--- a/Offer.sol
+++ b/Offer.sol
@@ -179,7 +179,7 @@ contract Offer {
     }
 
     // Once a proposal is submitted, the Contractor should call this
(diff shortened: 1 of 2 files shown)


Offer.sol
```diff
@@ -173,6 +178,15 @@ contract Offer {
         isContractValid = true;
     }
 
+    // Once a proposal is submitted, the Contractor should call this
+    // function to get the voting deadline of the proposal
```
colm: should be *set* the voting deadline

PR #242, opened 2016-06-12
Protect against recursive withdrawRewardFor attack

LefterisJP:
The DAO code is vulnerable to the [recursive call](http://vessenes.com/more-ethereum-attacks-race-to-empty-is-the-real-deal/) attack in the `withdrawRewardFor()` function. This fix should handle it for the next deployment of the DAO code.

A very big thank you to eththrowa from the [daohub forums](https://forum.daohub.org/t/bug-discovered-in-mkr-token-contract-also-affects-thedao-would-allow-users-to-steal-rewards-from-thedao-by-calling-recursively/4947) for spotting this.

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -744,9 +744,10 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
         reward = rewardAccount.balance < reward ? rewardAccount.balance : reward;
 
+        paidOut[_account] += reward;
         if (!rewardAccount.payOut(_account, reward))
             throw;
-        paidOut[_account] += reward;
+
         return true;
     }
 


glesaint:
So the issue was identified, the solution developed, but no one could implement it. How to setup DAOs to allow security hot-fixes?

ktorn:
Any comments on why this didn't get deployed urgently?
Especially taking into account that the attacker had already initiated his attack preparation 7 days in advance with the [split proposal](https://live.ether.camp/transaction/5798fbc45e3b63832abc4984b0f3574a13545f415dd672cd8540cd71f735db56).

LefterisJP:
@ktorn  The fix for the actual attack that happened is not yet in the v1.1 repo.
The reason that the fix in this PR or ... any of the v1.1 fixes did not make it into the DAO is simple. The DAO is a completely decentralized entity out of anyone's control.

The upgrade mechanism is slow and involves a vote of the token holders which would take at least 2 weeks.

DAO.sol
```diff
@@ -744,9 +744,10 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
         reward = rewardAccount.balance < reward ? rewardAccount.balance : reward;
 
+        paidOut[_account] += reward;
         if (!rewardAccount.payOut(_account, reward))
             throw;
```
ithinuel: If the payOut failed, should'nt it be removed from `paidOut[_account]` ?

LefterisJP: If it failed, the transaction would throw, so this would be as if it never happened.

philsong:
回顾一下，像看大片惊心动魄。

lucadonnoh (commented):
lgtm

PR #244, opened 2016-06-13
remove ExtraBalance

CJentzsch:


colm:
its tricky leaving the `creationPeriod` and the `refund` option in `TokenCreation.sol` this means the new DAO will be locked for 1 month while this time period elapses. It also means that the old DAO would have the option to ask for a refund.

PR #247, opened 2016-06-13
allow no-voters to withdraw

CJentzsch:
This is an alternative to https://github.com/slockit/DAO/pull/248 (only one should be merged)

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -326,7 +326,7 @@ contract DAOInterface {
 
     /// @param _account The address of the account which is checked.
     /// @return Whether the account is blocked (not allowed to transfer tokens) or not.
-    function isBlocked(address _account) internal returns (bool);
+    function getOrModifyBlocked(address _account) internal returns (bool);
 
     /// @notice If the caller is blocked by a proposal whose voting deadline
     /// has exprired then unblock him.
@@ -512,6 +512,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
             blocked[msg.sender] = _proposalID;
         }
 
+        // yay votes are blocked from splitting/withdrawing from the DAO
         if (_supportsProposal) {
             if (splitBlocked[msg.sender] == 0) {
                 splitBlocked[msg.sender] = _proposalID;
@@ -630,7 +631,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
         // don't allow withdrawing when token holder is blocked due to a vote
         // unVoteAll() would be an alternative
-        if (isSplitBlocked(msg.sender))
+        if (getOrModifySplitBlocked(msg.sender))
             throw;
 
         // Move ether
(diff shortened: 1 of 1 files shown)


DAO.sol
```diff
@@ -615,6 +626,47 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+        // unVoteAll() would be an alternative
+        if (isSplitBlocked(msg.sender))
+            throw;
+
+        // Move ether
+        uint fundsToBeMoved =
+            (balances[msg.sender] * actualBalance()) /
+            totalSupply;
+
+        msg.sender.call.value(fundsToBeMoved);
```
LefterisJP: **DANGER:** Here we are gonna become vulnerable to the recursive call exploit again! 

1. Change it to `send()`
2. Make all changes to the "state of the contract" before the actual send/call to protect from any recursive calls.

DAO.sol
```diff
@@ -902,16 +954,28 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
-        if (now > p.votingDeadline) {
+        if (p.open) {
             blocked[_account] = 0;
             return false;
         } else {
             return true;
         }
     }
 
+    function isSplitBlocked(address _account) internal returns (bool) {
```
LefterisJP: Please change the name of this function here and wherever it is used. The function may also change the `splitBlocked[]` array. So a better name would be: `getOrModifySplitBlocked()`

DAO.sol
```diff
@@ -902,16 +954,28 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+            splitBlocked[_account] = 0;
+            return false;
+        } else {
+            return true;
+        }
+    }
+
     function unblockMe() returns (bool) {
-        return isBlocked(msg.sender);
+        return isBlocked(msg.sender) && isSplitBlocked(msg.sender);
```
LefterisJP: This is something was really bothering me with the v1.0 DAO. Please as part of this PR rename `isBlocked` to `getOrModifyBlocked()`

DAO.sol
```diff
@@ -615,6 +627,45 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
+        // don't allow withdrawing when token holder is blocked due to a vote
+        // unVoteAll() would be an alternative
+        if (getOrModifySplitBlocked(msg.sender))
+            throw;
+
+        // Move ether
+        uint senderBalance = balances[msg.sender];
+        uint fundsToBeMoved = (senderBalance * actualBalance()) / totalSupply;
+        balances[msg.sender] = 0;
+        msg.sender.send(fundsToBeMoved);
```
colm: possible stack depth attack here. You need to wrap in if loop. I have fix but cannot push it to thedao repo.

DAO.sol
```diff
@@ -890,28 +941,40 @@ contract DAO is DAOInterface, Token, TokenCreation {
...
         } else {
             return true;
         }
     }
 
+    function getOrModifySplitBlocked(address _account) internal returns (bool) {
+        if (splitBlocked[_account] == 0)
+            return false;
+        Proposal p = proposals[splitBlocked[_account]];
+        if (p.open) {
```
colm: the token should stay blocked until the `splitExecutionPeriod` has passed.

LefterisJP: `splitExecutionPeriod` is 27 days as can be seen [here](https://github.com/slockit/DAO/blob/master/DAO.sol#L37). Are you sure you mean that? Perhaps you mean the `splitGracePeriod` from [this](https://github.com/slockit/DAO/pull/246/files#diff-b30f9d3be2a832949bdd72e23956c130R44) PR.

If it's the latter `p.open` stays true until `executeProposal()` is succesfull, right? It's kind of the same check, only relying on someone having succesfully called `executeProposal()`. In the case of a faulty proposal which always throws, someone can always call `executeProposal()` to have it closed after the `waitingPeriod()` is over.

PR #248, opened 2016-06-13
add withdraw and allow to change vote or unvote

CJentzsch:
This also allows for transferring tokens during the voting process.

This is an alternative to https://github.com/slockit/DAO/pull/247  (only one should be merged)

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -85,6 +85,9 @@ contract DAOInterface {
     mapping (address => uint) public blocked;
     mapping (address => uint) public splitBlocked;
 
+    // Map of addresses and proposal voted on by this address
+    mapping (address => uint[]) public votingRegister;
+
     // The minimum deposit (in wei) required to submit any proposal that is not
     // requesting a new Curator (no deposit is required for splits)
     uint public proposalDeposit;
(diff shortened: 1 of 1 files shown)


colm:
I think there is a game theory bug here. If a proposal looks like it will pass then the "nay" voters will unvote and withdraw. That means that there `nay` bias will collapse upon `yea` taking the lead. Having a grace period between execute and voting should be solve this.

jbaylina:
Imagine this sequence.
1 - Vote Yes with account 1 that has 1 Token
2 - Transfer 100 tokens from account 2 to account 1
3 - unvote from account 1.  

I believe that this will remove 100 yes and now I can continue voting.

4 - vote no with account 1 that now has 101 tokens.

colm:
Yes I thought this too. But `[ !isBlocked(_to)`](https://github.com/slockit/DAO/blob/develop/DAO.sol#L759) prevents the transfer to an address that has already voted. I need to confirm that all other transfer methods are blocked also.

colm:
one use case of hte `unVote`fuction is to allow to remove votes on proposals when `unVoteAll` will use to much gas.

This does not work becaues `unVote` does not change `votingRegister` which is iterated over in `unVoteAll` this means that in some cases `unVoteAll` will never be callable due to the block gas limit. 

This in turns means that `withdraw` will fail due to the same gas limit. Because it calls `unVoteAll`

PR #250, opened 2016-06-13
add swapTokens function

CJentzsch:
This will move the old DAO tokens to the new one and create new DAO tokens for the sender of the transaction.
This also enables the new DAO to retrieve the extraBalance.

### DIFF
--- a/DAO.sol
+++ b/DAO.sol
@@ -414,9 +414,15 @@ contract DAO is DAOInterface, Token, TokenCreation {
             || ( // 0xa9059cbb: transfer(address,uint256)
                 (_transactionData[0] == 0xa9 && _transactionData[1] == 0x05
                  && _transactionData[2] == 0x9c && _transactionData[3] == 0xbb))
+            || ( // 0x4e10c3ee: transferWithoutReward(address,uint256)
+                (_transactionData[0] == 0x4e && _transactionData[1] == 0x10
+                 && _transactionData[2] == 0xc3 && _transactionData[3] == 0xee))
             || ( // 0x23b872dd: transferFrom(address,address,uint256)
                 (_transactionData[0] == 0x23 && _transactionData[1] == 0xb8
                  && _transactionData[2] == 0x72 && _transactionData[3] == 0xdd))
+            || ( // 0xdbde1988: transferFromWithoutReward(address,address,uint256)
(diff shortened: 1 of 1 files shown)


LefterisJP:
@CJentzsch This PR can't be based on develop. Even Travis tells you that there is no `parentDAO` at the moment. It would be `privateCreation`. The way this PR is written it depends on https://github.com/slockit/DAO/pull/244

DAO.sol
```diff
@@ -855,6 +878,7 @@ contract DAO is DAOInterface, Token, TokenCreation {
 
     function minQuorum(uint _value) internal constant returns (uint _minQuorum) {
         // minimum of 20% and maximum of 53.33%
+        uint combinedTotalSupply = DAO(privateCreation).totalSupply() + totalSupply - tokensReceived;
```
colm: can we rename `privateCreation` to `parentDAO` or something. As it is pretty confusing.

DAO.sol
```diff
@@ -913,6 +937,21 @@ contract DAO is DAOInterface, Token, TokenCreation {
     function unblockMe() returns (bool) {
         return isBlocked(msg.sender);
     }
+
+    // approve DAO to transfer your tokens prior to that
+    function swapTokens() {
+         // privateCreation renamed to parentDAO removing extraBalance PR!
+        uint balance = DAO(privateCreation).balanceOf(msg.sender);
+        if (DAO(privateCreation).transferFrom(msg.sender, this, balance)) {
```
colm: `this` should be `this.address` ?

DAO.sol
```diff
@@ -401,6 +403,26 @@ contract DAO is DAOInterface, Token, TokenCreation {
         return true;
     }
 
+    function checkNewProposalRestrictions(
```
colm: Here we should include `transferFromWithoutReward` and `transferwithoutReward` to prevent 14% theft of the old DAO.

DAO.sol
LefterisJP: this will be done when it's combined with [this](https://github.com/slockit/DAO/pull/244) PR

DAO.sol
LefterisJP: You are correct. Very nice catch colm. I will add those now.

PR #260, opened 2016-06-16
fixed stack depth attack in withdrawAndUnvote

colm:


colm:
@LefterisJP there seems to be something strange with the compare here? The compare shows additions that are already in the base branch.

LefterisJP:
@colm I don't see anything strange with the diff. Can you specify what looks strange?

colm:
in my changes ln 683 -> 695 seem to be newly added by me. However these already exist in the base branch. I am worried about the integrity of the diffs. Because I use them in my reviews.

LefterisJP:
@colm The diff heuristic of git is not perfect. If you look carefully you will notice it's also removed code both above and below the lines you mentions and then readded it. This happens since you essentialy moved it and changed indentation.

If you use git locally for diff viewing and you use git version >= 2.9 you can enable `--compaction-heuristic` which as you can see [here](https://github.com/blog/2188-git-2-9-has-been-released) is quite smarter and shows the diffs in a much better way in situations like this.

PR #265, opened 2016-06-18
Update README.md

ramvi:
Inform people not to use in production until bug is fixed


CJentzsch:
better late then never ;-)

CJentzsch:
sorry, ment to merge it, but original repo doesnt exist anymore

PR #268, opened 2016-06-23
Adds a refund contract

alexvandesande:
This refund contract buys back tokens for how much it is able to secure and replaces them by a new token which can be used for future refunds.

### DIFF
--- a/refund.sol
+++ b/refund.sol
@@ -1,0 +1,91 @@
+/*
+
+CODE TO MAKE THE DAO INTO AN AUTOMATIC REFUND
+
+*/
+// A generic token contract
+contract token { 
+    mapping (address => uint256) public balanceOf;  
+    function transferFrom(address _from, address _to, uint256 _value) returns (bool success);
+    function mintToken(address target, uint256 mintedAmount);
+    uint256 public totalSupply;
+}
+
+
+/* define 'owned' */
+contract owned {
+    address public owner;
+
+    function owned() {
+        owner = msg.sender;
+    }
+
+    modifier onlyOwner {
+        if (msg.sender != owner) throw;
+        _
+    }
+
+    function transferOwnership(address newOwner) onlyOwner {
+        owner = newOwner;
+    }
+}
+
+contract DAORefund is owned {
+    // address of original DAO
+    token DAOTokens;
+    
+    // Address of new refund token (can be used to claim any additional eventual refund in the future)
+    token DAOBadgeOfHonor;
+    
+    // the price it will refund it for
+    uint public weiRefundedPerToken;
+
+    /* First time setup */
+    function DAORefund(address DAOAddress, address BOHAddress) {
+        DAOTokens = token(DAOAddress);
+        DAOBadgeOfHonor = token(BOHAddress);
+    }
+    
+    // to be called by the owner to lock the price so it guarantees all tokens are refunded the same
+    function setPrice() onlyOwner {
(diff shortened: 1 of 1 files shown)


refund.sol
```diff
@@ -0,0 +1,91 @@
...
+    // address of original DAO
+    token DAOTokens;
+    
+    // Address of new refund token (can be used to claim any additional eventual refund in the future)
+    token DAOBadgeOfHonor;
+    
+    // the price it will refund it for
+    uint public weiRefundedPerToken;
+
+    /* First time setup */
```
LefterisJP: Please provide some info here in the comments regarding the arguments. Especially some extra info aout the  BOHAddress which I assume is the 'Badge Of Honor' token address.

refund.sol
```diff
@@ -0,0 +1,91 @@
...
+    function DAORefund(address DAOAddress, address BOHAddress) {
+        DAOTokens = token(DAOAddress);
+        DAOBadgeOfHonor = token(BOHAddress);
+    }
+    
+    // to be called by the owner to lock the price so it guarantees all tokens are refunded the same
+    function setPrice() onlyOwner {
+        uint tokenSupply = DAOTokens.totalSupply() - DAOTokens.balanceOf(this);
```
LefterisJP: We will probably have to think a lot about the total supply here. Remember that the exploit allowed for people to create as many tokens as they want into a child DAO, stealing ether in the process and never burning any mother DAO tokens, thus not reducing the total supply of the mother DAO.

What about people who may have already gotten into a Dark DAO and are not malicious? We should probably add some kind of exception for them here.

I think that we should probably do some accounting, see who was naughty, see who may be an innocent victim in a dark DAO and then incorporate that accounting in here.

We should also consider blacklisting all known attackers from any refund. More money for the rest of the legitimate DTH.

refund.sol
```diff
@@ -0,0 +1,91 @@
...
+        weiRefundedPerToken = this.balance / tokenSupply;
+    }
+    
+    // allows tokens to be refunded by ether and badges of honors to be sent back
+    function convertTokens(uint amountOfTokens) {
+        if (weiRefundedPerToken == 0
+        || !DAOTokens.transferFrom(msg.sender, address(this), amountOfTokens))
```
LefterisJP: + 1 level of indentation here

refund.sol
```diff
@@ -0,0 +1,91 @@
...
+        if (weiRefundedPerToken == 0
+        || !DAOTokens.transferFrom(msg.sender, address(this), amountOfTokens)) 
+            throw;
+        
+        // do these last to avoid a recursion attack
+        DAOBadgeOfHonor.mintToken(msg.sender, amountOfTokens);
+        msg.sender.send(amountOfTokens * weiRefundedPerToken);
+    }
+
+    // Allows the refund contract on doing any actions except token transfers
```
LefterisJP: on doing -> to do

refund.sol
```diff
@@ -0,0 +1,91 @@
...
+    function execute(address _target, bytes _data) onlyOwner  {
+        // Don't allow it to make a 'approve' (0x095ea7b3) action on the dao 
+        if (_target == address(DAOTokens)
```
LefterisJP: We need some more exceptions here. I made a PR to your PR: https://github.com/alexvandesande/DAO/pull/1

CJentzsch:
cleaning up

PR #270, opened 2016-07-05
Refund child in seperate contract

CJentzsch:


chriseth:
The withdraw function should have a guard so that it cannot be called before the hard fork block number.

refund.sol
```diff
@@ -17,96 +17,22 @@ along with the DAO.  If not, see <http://www.gnu.org/licenses/>.
...
-        withdraw(_child);
-    }
-
-    function withdraw(){
-        withdraw(mother);
-    }
+    function withdraw(DAO _dao) {
+        if (block.number < 0xfffff) throw; // to be replaced by HF block number
+        if (!refundChild) {
+            refundChild = true;
```
chriseth: I don't think there is a need for the flag, we can use the address for both.

refund.sol
```diff
@@ -17,96 +17,22 @@ along with the DAO.  If not, see <http://www.gnu.org/licenses/>.
...
+            childRefund = new ChildRefund()
+            if (!childRefund.send(174558484783966581114679) || address(childRefund) == 0) throw;
```
chriseth: Is the number fixed? Is it "initial dao - dark dao - white dao"?

CJentzsch: yes, basically all ETH that is in all childDAO, minus the attacker (white and black) daos.

refund.sol
```diff
@@ -17,96 +17,20 @@ along with the DAO.  If not, see <http://www.gnu.org/licenses/>.
...
-        withdraw(_child);
-    }
-
-    function withdraw(){
-        withdraw(mother);
-    }
+    function withdraw(DAO _dao) {
+        if (block.number < 0xfffff) throw; // to be replaced by HF block number
+        if (address(childRefund) == 0)) {
+            childRefund = new ChildRefund()
```
chriseth: Missing `;`

CJentzsch: fixed

Sulfurtank:
bur ether and dont refund anyone, do the right thing scumbags

Sulfurtank:
burn eth and dont refund anyone do the right thing scumbags

PR #272, opened 2016-07-06
Refund contract fixes

LefterisJP:


refund.sol
```diff
@@ -109,7 +109,7 @@ contract Refund {
     function withdraw(DAO _dao) internal {
         uint balance = _dao.balanceOf(msg.sender);
 
-        if (!_dao.transferFrom(msg.sender, this, balance) || !msg.sender.send(balance * totalWeiSupply / totalSupply))
+        if (!_dao.transfer(this, balance) || !msg.sender.send(balance * totalWeiSupply / totalSupply))
```
CJentzsch: it needs to be `transferFrom` (same for the other examples)

LefterisJP: transferFrom(msg.sender, X, Y) is the same as transfer(X, Y). Plus will throw unless you have the approve. Perhaps there is something I am missing?

PR #273, opened 2016-07-16
Make sure enought gas and Fix Transaction-Ordering Dependence for refund()

u2:
I see most time we use `call.value()`, not `send`.

> `call.value()` (and using related functions like .callcode, .delegatecall) will provide the fallback function with virtually[1] as much gas as the caller has remaining.

We use `call.value()` the only reason is that it can secure enough gas, right?

### DIFF
--- a/TokenCreation.sol
+++ b/TokenCreation.sol
@@ -97,7 +97,6 @@ contract TokenCreation is TokenCreationInterface, Token {
         name = _tokenName;
         symbol = _tokenSymbol;
         decimals = _decimalPlaces;
-        
     }
 
     function createTokenProxy(address _tokenHolder) returns (bool success) {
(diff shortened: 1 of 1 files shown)


TokenCreation.sol
```diff
@@ -126,7 +126,7 @@ contract TokenCreation is TokenCreationInterface, Token {
                 extraBalance.payOut(address(this), extraBalance.accumulatedInput());
 
             // Execute refund
-            if (msg.sender.send(weiGiven[msg.sender])) { 
+            if (msg.sender.call.value(weiGiven[msg.sender])()) {
```
CJentzsch: with this you open up this function to the reentry attack

u2: I have updated. And Fix Transaction-Ordering Dependence, now it can avoid the reentry attack

u2: I have updated. And Fix Transaction-Ordering Dependence, now it can avoid the reentry attack.

u2: I have added `(balances[msg.sender] > 0 && weiGiven[msg.sender] > 0)` as pre-condition.

This is something like
```
    modifier onlyTokenholders {
            _
    }
```
Before, I moved this modifier from `dao.sol` to `token.sol`, and added the modifier for the `refund()`, but the test would fail, so I changed it back.

CJentzsch:
clean up

